PR #338, opened 2012-03-14
Translatable themes

gyuris:
Add translation support for themes to core.


gyuris:
@winhamwr, what was the problem with this PR?

PR #570, opened 2014-08-19
Parser to strip more than two consecutive line-breaks

mightyiam:
Replaces #451.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ WYMeditor.
 * [#546](https://github.com/wymeditor/wymeditor/pull/546)-Clicking on any UI
   element would cause loss of focus in non-IE browsers. Changing main
   containers would cause loss of selection.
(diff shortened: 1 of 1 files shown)


src/wymeditor/parser/xhtml-sax-listener.js
```diff
@@ -415,7 +417,17 @@ WYMeditor.XhtmlSaxListener.prototype.closeBlockTag = function(tag) {
         this._insideLI = false;
     }
 
-    this.output = this.output.replace(/<br \/>$/, '') +
+    // If a `br` is the only tag within the element that is being closed,
+    // remove that `br` tag.
+    this.output = this.output.replace(
+        new RegExp(
+            this._lastAddedOpeningTagString + '<br />$',
```
winhamwr: If this has characters that need to be escaped in a regex, we can get a SyntaxError.

PR #587, opened 2014-09-02
WYMeditor.Helper to jQuery extensions

mightyiam:
- Creates `src/wymeditor/jquery-extensions.js`
- Moves all contents of `WYMeditor.Helper` to `jQuery` in this file.
- Update all calls to those.
- Some were simply removed because they were not in use.


mightyiam:
Now that I look at it, it doesn't mean much to me, this PR.

PR #624, opened 2014-11-23
Combokeys

mightyiam:


CHANGELOG.md
```diff
@@ -23,6 +23,16 @@ WYMeditor.
 * [#648](https://github.com/wymeditor/wymeditor/pull/647),
   [#641](https://github.com/wymeditor/wymeditor/pull/641)
   Object resizing (images, primarily) in Firefox and IE has been disabled.
+* [#639](https://github.com/wymeditor/wymeditor/pull/639)
```
mightyiam: This is from #639

CHANGELOG.md
```diff
@@ -32,8 +42,12 @@ WYMeditor.
   New `editor.doesElementContainSelection` method.
 * [#512](https://github.com/wymeditor/wymeditor/pull/512)
   New undo/redo methods in the API.
+* [#624](https://github.com/wymeditor/wymeditor/pull/624)
+  New `editor.keyboard.combokeys` for managing keyboard shortcuts.
 * [#641](https://github.com/wymeditor/wymeditor/pull/641)
   New `editor.getSelectedImage()`
+* [#639](https://github.com/wymeditor/wymeditor/pull/639)
+  New dialogs API. See API documentation for the `wym.dialog` method.
```
mightyiam: From #639 .

CHANGELOG.md
```diff
@@ -52,6 +66,8 @@ WYMeditor.
 
   Undo/redo, even if unused, may eat a lot of RAM, depending on the document
   size and the browser. To be optimized.
+* [#624](https://github.com/wymeditor/wymeditor/pull/624)
+  Keyboard shortcuts are more consistent cross-platform.
```
mightyiam: They are.

Gruntfile.js
```diff
@@ -248,6 +252,7 @@ module.exports = function (grunt) {
                 src: [
                     "<%= yeoman.app %>/lib/jquery.browser.js",
                     "<%= yeoman.app %>/wymeditor/core.js",
+                    "<%= yeoman.app %>/wymeditor/editor/dialogs.js",
```
mightyiam: From #639 .

Gruntfile.js
```diff
@@ -511,7 +517,8 @@ module.exports = function (grunt) {
 
         grunt.task.run([
             'bower',
-            'browserify:objectHistory',
+            'browserify:externalModules',
+            'browserify:keysim',
```
mightyiam: `objectHistory` is in `externalModules` now.
`keysim` is not part of that because it is only used for testing.

Gruntfile.js
```diff
@@ -521,7 +528,8 @@ module.exports = function (grunt) {
 
     grunt.registerTask('test', [
         'bower',
-        'browserify:objectHistory',
+        'browserify:externalModules',
+        'browserify:keysim',
```
mightyiam: See above.

Gruntfile.js
```diff
@@ -532,7 +540,7 @@ module.exports = function (grunt) {
         'useminPrepare',
         'bower-install-simple',
         'bower-linker:dist-examples',
-        'browserify:objectHistory',
+        'browserify:externalModules',
```
mightyiam: See above.

docs/customizing_wymeditor/index.rst
```diff
@@ -394,9 +394,6 @@ or useful for deep customization.
```
mightyiam: All changes in this file are from #639

docs/writing_plugins/api.rst
```diff
@@ -364,21 +364,10 @@ Undo
     undo an action.
```
mightyiam: All changes in this file are from #639

src/examples/14-more-inline-elements.html
```diff
@@ -1,160 +0,0 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
mightyiam: Removed in #639 .

src/test/acronym-dialog.html
```diff
@@ -1,154 +0,0 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
mightyiam: Removed in #639

src/test/load-src.dev.js
```diff
@@ -18,14 +18,16 @@ function loadWymSrc(srcPath, extraRequirements, jqueryVersion) {
             srcPath + 'wymeditor/rangy/rangy-core.js',
             srcPath + 'wymeditor/rangy/rangy-selectionsaverestore.js',
             srcPath + 'wymeditor/core.js',
+            srcPath + 'wymeditor/editor/dialogs.js',
```
mightyiam: From #639

src/test/unit/index.html
```diff
@@ -65,6 +68,7 @@
             "specific_feature_tests/images.js",
             "specific_feature_tests/class_toggling.js",
             "specific_feature_tests/links.js",
+            "specific_feature_tests/dialogs.js",
```
mightyiam: From #639

src/test/unit/keysim-globalifier.js
```diff
@@ -0,0 +1,15 @@
...
+   ) {
+    // This browser does not properly implement
+    // `Object.prototype.defineProperty`.
+    // This is required by the external keyboard event simulation module,
+    // `keysim`.
+    window.skipKeyboardShortcutTests = true;
+} else {
+    window.Keysim = require("keysim");
+    window.skipKeyboardShortcutTests = false;
+}
```
mightyiam: Some [progress was made](https://github.com/eventualbuddha/keysim.js/issues/7) in making Keysim work in IE7&8. IE8 compatibility seems close. Should we consider it blocking?

winhamwr: I would be OK with creating a separate issue to add IE8 testing support and merging this in. Doesn't make sense to block getting user feedback at this point.

src/test/unit/manipulation-test-helper.js
```diff
@@ -31,6 +33,9 @@
  *     `manipulationFunc`
  *         Optional; The manipulation function to be tested. Receives one
  *         argument, the WYMeditor instance.
+ *     `manipulationKeyCombo`
+ *         Optional; A key combination that is expected to trigger the
+ *         manipulation. For example, "ctrl+b".
```
mightyiam: Added this to `manipulationTestHelper`.

src/test/unit/manipulation-test-helper.js
```diff
@@ -124,6 +131,10 @@ function manipulationTestHelper(a) {
         executions.push(EXECUTE.UI_CLICK);
     }
 
+    if (typeof a.manipulationKeyCombo === "string") {
+        executions.push(EXECUTE.KEY_COMBO);
```
mightyiam: I've moved the check for `skipKeyboardShortcutTests` from here to inside `manipulateAndAssert`, below. It fixed a bug.

src/test/unit/manipulation-test-helper.js
```diff
@@ -145,6 +156,16 @@ function manipulationTestHelper(a) {
 
     function manipulateAndAssert(manipulationCause) {
 
+        if (
+            skipKeyboardShortcutTests &&
+            manipulationCause === EXECUTE.KEY_COMBO
+        ) {
+            if (expect() === null) {
+                expect(0);
```
mightyiam: These `expect` calls will have to be changed again, after #659 is merged in.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
  */
 /* jshint latedef: nofunc */
 function dialogTestHelper(args) {
-    // Sinon doesn't allow wrapping of `window.open` with a spy because `open`
-    // doesn't look like a function in these browsers.
-    var sinonCantWrapWindowOpen = jQuery.browser.msie &&
-        jQuery.browser.versionNumber <= 8;
```
mightyiam: Too much complication with wrapping of this method. Don't really need to spy on it. Other assertions are sufficient.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
...
+        prepareFunc: args.prepareFunc,
         manipulationClickSelector:
             getDialogToolbarSelector(args.dialogName),
+        manipulationKeyCombo:
+            getDialogKeyCombo(args.dialogName),
         expectedResultHtml: args.noChangeHtml,
         additionalAssertionsFunc: additionalAssertionsFunc
     });
 
-    function prepareFunc(wymeditor) {
```
mightyiam: Removed this `prepareFunc` wrapper.

mightyiam: Because the only thing it was doing is wrapping `window.open`.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
...
-            // The dialog function is called, whether the dialog is to be
-            // opened or not. It has the code that determines that.
-            wymeditor.dialog.calledOnce,
-            "Dialog function called once"
-        );
-
         // Its return value is either the dialog window or `false`.
-        dialogWindowOrFalse = wymeditor.dialog.returnValues[0];
+        var dialogReturns = wymeditor.dialog.returnValues,
+            dialogWindowOrFalse = dialogReturns[dialogReturns.length - 1];
```
mightyiam: `additionalAssertionsFunc` for the dialogs tests is to get the return value of the wrapped `dialog` method, which is either the dialog window or `false`.

The change here is that now we have multiple return values, because one of the dialogs (link) opens using a key combo, as well.

So for each `manipulationCause` we get the return value (the last array member).

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
...
             getDialogToolbarSelector(args.dialogName),
+        manipulationKeyCombo:
+            getDialogKeyCombo(args.dialogName),
         expectedResultHtml: args.noChangeHtml,
         additionalAssertionsFunc: additionalAssertionsFunc
     });
 
-    function prepareFunc(wymeditor) {
-        // Wrap the dialog function with a spy
-        args.currentTest.spy(wymeditor, "dialog");
```
mightyiam: Instead of wrapping the dialog here and unwrapping it afterwards, with each `manipulationCause` run, we wrap it above, in the beginning of `dialogTestHelper`.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
...
     function additionalAssertionsFunc(wymeditor) {
-        var dialogWindowOrFalse;
-        expect(expect() + 1);
-
-        ok(
-            // The dialog function is called, whether the dialog is to be
-            // opened or not. It has the code that determines that.
-            wymeditor.dialog.calledOnce,
-            "Dialog function called once"
-        );
```
mightyiam: Checking that it is called is redundant.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -107,13 +86,6 @@ function dialogTestHelper(args) {
 
     function assertOpened(dialogWindow) {
         expect(expect() + 2);
-        if (sinonCantWrapWindowOpen !== true) {
-            expect(expect() + 1);
-            ok(
-                window.open.calledOnce,
-                "`window.open` was called"
-            );
-        }
```
mightyiam: As mentioned above, this check is redundant.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -127,13 +99,6 @@ function dialogTestHelper(args) {
     }
     function assertNotOpened(returnedFalse) {
         expect(expect() + 1);
-        if (sinonCantWrapWindowOpen !== true) {
-            expect(expect() + 1);
-            ok(
-                window.open.callCount === 0,
-                "`window.open` was not called"
-            );
-        }
```
mightyiam: As mentioned above, this check is redundant.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -154,6 +119,12 @@ function getDialogToolbarSelector(dialogName) {
     return ".wym_section.wym_tools.wym_buttons li a[name=" + dialogName + "]";
 }
 
+function getDialogKeyCombo(dialogName) {
+    return {
+        CreateLink: "ctrl+k"
+    }[dialogName] || null;
+}
```
mightyiam: This returns `null` if the `dialogName` is not in this literal object expression.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
...
 
     manipulationTestHelper({
         startHtml: args.noChangeHtml,
         setCaretInSelector: args.setCaretInSelector,
-        prepareFunc: prepareFunc,
+        prepareFunc: args.prepareFunc,
         manipulationClickSelector:
             getDialogToolbarSelector(args.dialogName),
+        manipulationKeyCombo:
+            getDialogKeyCombo(args.dialogName),
```
mightyiam: If a certain dialog isn't supposed to have a key combo, `null` is returned from `getDialogKeyCombo` and `manipultionTestHelper` only uses the `manipulationClickSelector`.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -173,266 +144,235 @@ function getDialogDocumentTitle(dialogName) {
     return TITLES[dialogName];
 }
 
-test("Link dialog doesn't open when no selection; by toolbar button click",
```
mightyiam: Removed `by toolbar button click` from the names of all of these tests because we now test not only by toolbar button click, but also by key combo (in the case of the link dialog).

src/wymeditor/editor/keyboard.js
```diff
@@ -0,0 +1,70 @@
...
+    {
+        combo: "mod+i",
+        cb: function () {
+            this.exec(WYMeditor.EXEC_COMMANDS.ITALIC);
+            // Prevents native action. Caution: not covered by tests.
+            return false;
+        }
+    },
+    {
+        combo: "mod+k",
```
mightyiam: Is this reasonable? Some Google products behave this way.

src/test/unit/specific_feature_tests/lists.js
```diff
@@ -2756,79 +2766,48 @@ test("Double indent correction", function () {
...
-
-    simulateKey(WYMeditor.KEY_CODE.TAB, actionElement);
-    wymEqual(wymeditor, expectedHtml);
+    manipulationTestHelper({
+        startHtml: nestedListHtml,
+        setCaretInSelector: "#li_7",
+        manipulationKeyCombo: "tab",
+        expectedResultHtml: li_7_indentedHtml,
+        testUndoRedo: true
+    });
```
mightyiam: Rewrote using `manipulationTestHelper`.

src/test/unit/specific_feature_tests/lists.js
```diff
@@ -2756,79 +2766,48 @@ test("Double indent correction", function () {
...
+        setCaretInSelector: "#li_1_1",
+        manipulationKeyCombo: "shift+tab",
+        expectedResultHtml: [""
+            , "<ol>"
+                , "<li id=\"li_1_1\">1_1</li>"
+                , "<li id=\"li_2\">2</li>"
+            , "</ol>"
+        ].join(""),
+        testUndoRedo: true
+    });
```
mightyiam: Rewrote using `manipulationTestHelper`.

src/test/unit/specific_feature_tests/lists.js
```diff
@@ -2756,79 +2766,48 @@ test("Double indent correction", function () {
...
-    simulateKey(WYMeditor.KEY_CODE.TAB, actionElement);
-    wymEqual(wymeditor, expectedHtml);
+    var noChangeHtml = "<p id=\"p_1\">test</p>";
+    manipulationTestHelper({
+        startHtml: noChangeHtml,
+        setCaretInSelector: "#p_1",
+        manipulationKeyCombo: "tab",
+        expectedResultHtml: noChangeHtml,
+        testUndoRedo: true
+    });
```
mightyiam: Rewrote using `manipulationTestHelper`.

src/wymeditor/editor/trident-7.js
```diff
@@ -48,9 +48,6 @@ WYMeditor.WymClassTrident7.prototype.rawHtml = function (html) {
 WYMeditor.WymClassTrident7.prototype._docEventQuirks = function () {
     var wym = this;
 
-    jQuery(wym._doc).bind("keydown", function (evt) {
-        wym._keydown(evt);
-    });
```
mightyiam: This shouldn't have been here, anyway, I think.

src/wymeditor/editor/undo-redo.js
```diff
@@ -23,6 +25,23 @@ WYMeditor.UndoRedo = function (wym) {
...
+    wym.keyboard.combokeys.bind(
+        "mod+z",
+        function () {
+            wym.undoRedo.undo();
+            // Prevents native action. Caution: not covered by tests.
+            return false;
+        }
+    );
+    wym.keyboard.combokeys.bind(
+        jQuery.browser.mac ? "shift+meta+z" : "ctrl+y",
```
mightyiam: I'm interested to get this tested on a mac. Do you think that this is correct, @winhamwr ?

src/test/unit/manipulation-test-helper.js
mightyiam: @winhamwr so please merge #659 first, then review this and don't merge—let me fix this and then I'll merge.

Gruntfile.js
```diff
@@ -193,13 +193,17 @@ module.exports = function (grunt) {
                     debug: true
                 }
             },
-            objectHistory: {
+            externalModules: {
```
winhamwr: :+1:

docs/writing_plugins/api.rst
```diff
@@ -677,6 +677,20 @@ Example:
...
+
+For each instance of the editor, a Combokeys instance is instantiated, attached
+to the document and assigned as ``wym.keyboard.combokeys``.
+
+This Combokeys instance can be used to attach keyboard shortcuts.
+
+Please refer to `Combokeys`_' documentation.
+
+.. _Combokeys: http://github.com/mightyiam/combokeys
+
```
winhamwr: :+1:

src/test/unit/manipulation-test-helper.js
winhamwr: :+1:

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -41,48 +41,27 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
  */
 /* jshint latedef: nofunc */
 function dialogTestHelper(args) {
-    // Sinon doesn't allow wrapping of `window.open` with a spy because `open`
-    // doesn't look like a function in these browsers.
-    var sinonCantWrapWindowOpen = jQuery.browser.msie &&
-        jQuery.browser.versionNumber <= 8;
+    var wymeditor = jQuery.wymeditors(0);
```
winhamwr: How about `tryDialogWithClickAndKeyboardThenAssert`? Long name, but this function does a lot.

Mostly, it's very unclear from reading these tests right now that we're testing both buttons and toolbar buttons. The docstring could also be updated to say that "...window was opened or not via both a toolbar click and a keyboard shortcut".

src/test/unit/specific_feature_tests/lists.js
winhamwr: :+1:

src/wymeditor/editor/keyboard.js
winhamwr: :+1: 

We need to figure out a good method for exposing our keyboard shortcuts to users, though.

src/wymeditor/editor/undo-redo.js
winhamwr: Would it be possible to bind both, on macs? It seems like MS Word for Mac OSX still uses ctrl + y, so some users will be familiar with that.

src/wymeditor/editor/webkit.js
```diff
@@ -11,38 +11,11 @@ WYMeditor.WymClassWebKit = function (wym) {
...
-        if (evt.which === WYMeditor.KEY_CODE.I) {
-            //CTRL+i => EMPHASIS
-            wym._exec(WYMeditor.EXEC_COMMANDS.ITALIC);
-            evt.preventDefault();
-        }
-    } else if (evt.shiftKey && evt.which === WYMeditor.KEY_CODE.ENTER) {
-        // Safari 4 and earlier would show a proper linebreak in the editor and
-        // then strip it upon save with the default action in the case of
-        // inserting a new line after bold text
-        wym._exec('InsertLineBreak');
```
winhamwr: Why was this removed and not replaced?

src/wymeditor/iframe/default/wymiframe.html
```diff
@@ -21,5 +21,5 @@
 <title>WYMeditor iframe</title>
 <link rel="stylesheet" type="text/css" media="screen" href="wymiframe.css" />
 </head>
-<body class="wym_iframe"></body>
+<body></body>
```
winhamwr: Does this mean that existing iframes that already contain `wym_iframe` won't work with keyboard shortcuts? That's a pretty big backwards incompatible change. We should remove it (if it exists), register shortcuts and then re-add it, if possible.

PolicyStat uses a custom `iframe` that would be affected by this, for example. At the very least, it should be called out as a backwards incompatible change so that folks know to update their iframes.

src/wymeditor/plugins/list/jquery.wymeditor.list.js
```diff
@@ -27,35 +27,16 @@ ListPlugin.prototype.bindEvents = function() {
...
-ListPlugin.prototype.handleKeyDown = function(evt) {
-    var doc = this,
-        wym = WYMeditor.INSTANCES[doc.title],
-        listPlugin = wym.listPlugin,
-        container = wym.selectedContainer();
-        name = container.tagName.toLowerCase();
-    // We only care about tabs when we're inside a list
-    if (name != "li") {
-        return null;
-    }
```
winhamwr: You're sure of this?

src/wymeditor/plugins/list/jquery.wymeditor.list.js
```diff
@@ -27,35 +27,16 @@ ListPlugin.prototype.bindEvents = function() {
...
-            return false;
         }
-    }
-
-    return null;
+    );
+    wym.keyboard.combokeys.bind(
+        "shift+tab",
+        function () {
+            wym.exec(WYMeditor.EXEC_COMMANDS.OUTDENT);
```
winhamwr: And we don't need to short-circuit default behavior by returning false?

src/test/unit/specific_feature_tests/dialogs.js
mightyiam: I've updated the docstring. It seems enough to refrain from such a long name, to me.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -21,6 +21,9 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
  * This is a helper for testing whether a dialog window was opened or not.
  * It also tests whether the dialog has the correct title.
  *
+ * If the dialog window can be opened using a keyboard shortcut, add that to
+ * `getDialogKeyCombo` in this file, and that will be tested as well.
```
mightyiam: Does this help keep the name concise?

src/wymeditor/editor/keyboard.js
mightyiam: Right. De–facto standard is via tooltips on buttons. Aded complexity is displaying the correct ones according to the OS.

src/wymeditor/editor/undo-redo.js
```diff
@@ -23,6 +25,23 @@ WYMeditor.UndoRedo = function (wym) {
...
+    wym.keyboard.combokeys.bind(
+        "mod+z",
+        function () {
+            wym.undoRedo.undo();
+            // Prevents native action. Caution: not covered by tests.
+            return false;
+        }
+    );
+    wym.keyboard.combokeys.bind(
+        ["shift+meta+z", "ctrl+y"],
```
mightyiam: @winhamwr does this look good?

src/wymeditor/editor/webkit.js
mightyiam: You're right. I see that I require some more specific browser detection because Chrome and Safari work differently, here. Hence #662.

src/wymeditor/iframe/default/wymiframe.html
mightyiam: > Does this mean that existing iframes that already contain `wym_iframe` won't work with keyboard shortcuts?

Why won't they work with keyboard shortcuts?

src/wymeditor/editor/undo-redo.js
winhamwr: Should it be mod+y instead of ctrl+y? I think it looks good, otherwise.

src/test/unit/specific_feature_tests/dialogs.js
winhamwr: Docstrings and comments are very poor replacements for descriptive names, but are good icing on the cake. `dialogTestHelper` doesn't tell me anything at all about what the function *does*, which means even a cursory look through this module requires the reader to find the definition of `dialogTestHelper` and attempt to answer their own version of the "what does this do?". With `tryDialogWithClickAndKeyboardThenAssert`, I keep the abstraction while reading and can just continue on looking at test cases if I don't need to understand how that function tries the dialog with a click and then a keyboard shortcut and asserts.

There might be a better name, but I think the things needed conveying to understand what this function does:
* It does dialog things
* It uses a click
* It uses a keyboard shortcut
* It runs some assertions on the results

The long name is a sign that maybe we could have broken this up into multiple functions (or perhaps not), but lots of typing isn't a good reason to leave a generic name. This is currently the semantic equivalent of `doTestStuff` :)

src/wymeditor/iframe/default/wymiframe.html
winhamwr: With [this change](https://github.com/wymeditor/wymeditor/pull/624/files#r21754781), you modified the way `wym_iframe` is required and used. I'm trying to understand why you did that and how it will affect existing custom iframes.

mightyiam: That change means that the `wym_iframe` class will be added to the Iframe upon instantiation.

This is why I've removed it from the Iframe HTML—it is just redundant and means more code and one less thing you must do to have a working custom Iframe.

I don't see backward incompatibility.

src/wymeditor/editor/webkit.js
mightyiam: Here it is, right above. Also see test for it.

src/test/unit/test.js
```diff
@@ -93,6 +93,54 @@ test("Focusing on document that has designMode off, turns it on", function () {
     });
 });
 
+test("WebKit's shift+enter inserts a line break and prevents default " +
+    "action", function () {
```
mightyiam: New test for WebKit shift+enter

src/wymeditor/editor/webkit.js
mightyiam: I'm trying to install a hackintosh in my VirtualBox to test. Until then... Good luck Safari...

mightyiam: Midori, a WebKit browser does respond well to this, at least.

src/wymeditor/plugins/list/jquery.wymeditor.list.js
mightyiam: Indeed.

src/wymeditor/editor/keyboard.js
```diff
@@ -0,0 +1,83 @@
...
+        cb: function () {
+            var wym = this;
+            wym.dialog(WYMeditor.DIALOGS.CreateLink);
+            // Prevents native action. Caution: not covered by tests.
+            return false;
+        }
+    },
+    {
+        combo: "tab",
```
mightyiam: I've discovered this now. Bug fix!

src/wymeditor/plugins/list/jquery.wymeditor.list.js
```diff
@@ -27,35 +27,18 @@ ListPlugin.prototype.bindEvents = function() {
...
-        if (evt.shiftKey) {
-            wym.exec(WYMeditor.EXEC_COMMANDS.OUTDENT);
-            return false; // Short-circuit normal tab behavior
-        } else {
-            wym.exec(WYMeditor.EXEC_COMMANDS.INDENT);
+    wym.keyboard.combokeys.bind(
+        "tab",
+        function () {
+            wym.indent();
             return false;
```
mightyiam: @winhamwr you're right. Default action should be prevented here.

src/wymeditor/plugins/list/jquery.wymeditor.list.js
```diff
@@ -27,35 +27,18 @@ ListPlugin.prototype.bindEvents = function() {
...
         }
-    }
-
-    return null;
+    );
+    wym.keyboard.combokeys.bind(
+        "shift+tab",
+        function () {
+            wym.outdent();
```
mightyiam: Also changed to use `outdent`/`indent` directly, instead of through `exec`.

src/test/unit/specific_feature_tests/dialogs.js
mightyiam: The way I say it is that this helper is meant to be a general dialog test helper that does a bunch of things which are common throughout the testing of dialogs.

It may, in the future, be extended to do some more things, like, for example, support opening of dialogs by the `dialog` function itself, passing the dialog object. While that will still fit the essence of this helper, if we change it to `tryDialogWithClickAndKeyboardThenAssert`, we would be compelled to change that name again.

The name `dialogTestHelper` currently represents the essence of the helper, well, as I see it. Functional changes within the scope of a dialog test helper will not require renaming.

Now, whether the person who sees this function for the first time is more enlightened by `dialogTestHelper` or `tryDialogWithClickAndKeyboardThenAssert`, that is an interesting question.

It seems that `dialogTestHelper` describes the essence of the function and `try...ThenAssert` describes the form of the function.

Computers don't understand any essence. They are made of form and operate on form. We, on the other hand, are intelligent (even though we may not act like it at times)—we understand essence. Just a thought.

It seems that while naming this function, I have done that out of the concept that I held regarding its essence.

When you come and read along, is `dialogTestHelper` not good enough to help you understand that it helps test dialogs? Would you like the form level of its operation to be present as its name? What about possible future changes, like I mentioned?

Is it not better that the form–level explanations be kept in the docstring, allowing you to understand what this function does if you think you want to know more?

Perhaps we should reach a balance more towards the form, like `openDialogAndAssert`, which is likely to represent the core functionality of this helper, even if changes are made?

src/wymeditor/editor/undo-redo.js
```diff
@@ -23,6 +25,23 @@ WYMeditor.UndoRedo = function (wym) {
...
+    wym.keyboard.combokeys.bind(
+        "mod+z",
+        function () {
+            wym.undoRedo.undo();
+            // Prevents native action. Caution: not covered by tests.
+            return false;
+        }
+    );
+    wym.keyboard.combokeys.bind(
+        ["shift+meta+z", "mod+y"],
```
mightyiam: @winhamwr `mod+y`.

mightyiam: This isn't according to [their guidelines](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/Keyboard.html#//apple_ref/doc/uid/20000957-CH84-SW3) but if you say the MSOffice for Mac uses command+y...

CHANGELOG.md
```diff
@@ -64,6 +66,10 @@ WYMeditor.
 
   Undo/redo, even if unused, may eat a lot of RAM, depending on the document
   size and the browser. To be optimized.
+* [#624](https://github.com/wymeditor/wymeditor/pull/624)
+  * Keyboard shortcuts are more consistent cross-platform.
+  * Fix: pressing tab in WebKit and Blink creates span. caveat: now tab can’t
+    be used to focus on the next element (taking focus outside the document).
```
mightyiam: This entry about the tab.

src/wymeditor/editor/keyboard.js
winhamwr: Whoa. Nice find!

Does this fix break anything with the list plugin's tab handling? I'm not exactly sure how multiple combos bound to the same key are handled.

src/wymeditor/iframe/default/wymiframe.html
winhamwr: Gotcha. I wasn't sure if the presence of that class would break anything, but it sounds like it doesn't. :+1:

src/wymeditor/editor/undo-redo.js
winhamwr: It does. Mac OSX purists don't like it, but that's MS for you :P

src/wymeditor/editor/keyboard.js
mightyiam: Sequentially, IIRC. In any case, it shouldn't and doesn't interfere because it doesn't do anything manipulation.

src/test/unit/keysim-globalifier.js
winhamwr: Can you create this ticket and link to it in a comment before merging this in?

src/test/unit/specific_feature_tests/dialogs.js
winhamwr: > it may, in the future, be extended to do some more things

And at that time, I will probably recommend splitting it into another function, or at least a wrapper, so that the name can be useful. 

> When you come and read along, is dialogTestHelper not good enough to help you understand that it helps test dialogs?

Tests what about dialogs, though. That's the question you need to understand what the test is actually doing and where the work is being abstracted. The test names don't make it clear that the keyboard shortcuts and mouse clicks are both being tested. That's OK, as long as a read-through of the test code makes it clear (since the test code is concise).

> which is likely to represent the core functionality of this helper, even if changes are made?

Making name recommendations based mostly on the future is a dicey prospect. Future us will have a lot more information and ability to make a decision on a good name, so let's trust ourselves to make that decision. Current us has a function that does a lot, but that can be described in a way that the reader doesn't need to read the implementation to understand what's happening in the test. We should use that descriptive name.

docstrings are for the why and for the details applicable to someone who wants to use a function. As a reader, it's the name of a thing that is responsible for conveying what a function does.

I think you understand where I'm coming from, so this name is your call, now.

src/test/unit/test.js
winhamwr: :+1:

winhamwr:
Other than the IE8 testing issue creation and your decision on the dialog helper name, this looks ready to merge. I'm excited about that span fix. I've actually seen that when manipulating lists and I never actually figured out why it was happening.

src/test/unit/keysim-globalifier.js
mightyiam: Thanks. Done.

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -37,57 +44,37 @@ module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
...
+ *     DIALOG_EXPECT_NOT_OPENED` (These
  *     are constants).
  */
 /* jshint latedef: nofunc */
-function dialogTestHelper(args) {
-    // Sinon doesn't allow wrapping of `window.open` with a spy because `open`
-    // doesn't look like a function in these browsers.
-    var sinonCantWrapWindowOpen = jQuery.browser.msie &&
-        jQuery.browser.versionNumber <= 8;
+function tryDialogWithClickAndKeyboardThenAssert(args) {
```
mightyiam: @winhamwr alrighty

src/test/unit/specific_feature_tests/dialogs.js
```diff
@@ -17,10 +17,17 @@
 
 module("dialogs-opening_or_not", {setup: prepareUnitTestModule});
 
+var DIALOG_EXPECT_OPENED = "Expect this dialog to have been opened. " +
+    "Expect it with all of your heart.";
+var DIALOG_EXPECT_NOT_OPENED = "Nope. The dialog should not have opened.";
```
mightyiam: @winhamwr these are now not properties of `dialogTestHelper` because it was renamed to be much longer. They are globals now.

mightyiam:
Awesome. Thank you!

PR #658, opened 2015-01-23
Remove unused checked in qunit dist files

mightyiam:


winhamwr:
Looks like we forgot to remove these when we added the explicitly version-numbered files?

mightyiam:
> Looks like we forgot to remove these when we added the explicitly version-numbered files?

It wasn't me

PR #662, opened 2015-01-27
More specific browser detection

mightyiam:


mightyiam:
In #624 it seems that more specific browser detection is required.

Specifically, for detecting Chrome, Safari and WebKit, instead of treating them all as WebKit.

Gruntfile.js
```diff
@@ -254,6 +254,8 @@ module.exports = function (grunt) {
                         "document-structure-manager.js",
                     "<%= yeoman.app %>/wymeditor/editor/gecko.js",
                     "<%= yeoman.app %>/wymeditor/editor/webkit.js",
+                    "<%= yeoman.app %>/wymeditor/editor/blink.js",
+                    "<%= yeoman.app %>/wymeditor/editor/safari.js",
```
mightyiam: Concatenation for distribution

bower.json
```diff
@@ -14,7 +14,7 @@
   "devDependencies": {
     "requirejs": "~2.1",
     "jquery": "~1.4.4",
-    "jquery.browser": "~0.0.6",
+    "jquery.browser": "~0.0.7",
```
mightyiam: Officially newer version while we're at it.

src/test/load-src.dev.js
```diff
@@ -25,6 +25,8 @@ function loadWymSrc(srcPath, extraRequirements, jqueryVersion) {
             srcPath + 'wymeditor/editor/trident-pre-7.js',
             srcPath + 'wymeditor/editor/trident-7.js',
             srcPath + 'wymeditor/editor/webkit.js',
+            srcPath + 'wymeditor/editor/blink.js',
+            srcPath + 'wymeditor/editor/safari.js',
```
mightyiam: Load them in dev–env.

src/wymeditor/core.js
```diff
@@ -1140,3 +1140,39 @@ WYMeditor.Helper = {
...
+
+/**
+    WYMeditor._getWymClass
+    ======================
+
+    Returns the constructor for the browser-specific wymeditor instance.
+
+    If does not detect a supported browser, returns false;
+*/
+WYMeditor._getWymClass = function () {
```
mightyiam: The new browser detecting logic.

src/wymeditor/editor/base.js
```diff
@@ -33,29 +34,17 @@ WYMeditor.editor.prototype._init = function () {
...
-            // https://stackoverflow.com/questions/20655470
-            //WymClass = new WYMeditor.WymClassBlink(wym);
-            // For now we use the WebKit editor class in Blink.
-            WymClass = new WYMeditor.WymClassWebKit(wym);
-        } else {
-            WymClass = new WYMeditor.WymClassWebKit(wym);
-        }
-    }
+    // Get the constructor for the browser-specific instance
+    WymClass = WYMeditor._getWymClass();
```
mightyiam: Delegate to the new `WYMeditor._getWymClass`.

mightyiam: Also, notice that the constructor itself gets assigned to `WymClass` (justifies the capital W).

Only below it gets instantiated and assigned to the newly added `browserInstance`.

src/wymeditor/editor/base.js
```diff
@@ -33,29 +34,17 @@ WYMeditor.editor.prototype._init = function () {
...
 
-    if (WymClass === false) {
+    if (!WymClass) {
+        // We don't support this browser. Don't initialize.
```
winhamwr: It feels like that if we can't reliably print to the console via `console.warn` to give a developer some feedback, we should throw an exception. This silent failure would be really frustrating behavior for someone to debug.

src/wymeditor/editor/base.js
winhamwr: :+1: 
Good level of encapsulation.

For a name, how about `_getWymClassForBrowser` to make it clear that the class logic is about the user's web browser? Then, you don't actually need the comment. Better names > comments

src/wymeditor/core.js
```diff
@@ -1140,3 +1140,39 @@ WYMeditor.Helper = {
...
+        case "mozilla":
+            return WYMeditor.WymClassGecko;
+        case "chrome":
+            return WYMeditor.WymClassBlink;
+        case "safari":
+            return WYMeditor.WymClassSafari;
+    }
+
+    if (jQuery.browser.webkit) {
+        return WYMeditor.WymClassWebKit;
```
winhamwr: What case does this handle?

mightyiam: 1. https://en.wikipedia.org/wiki/Midori_%28web_browser%29
1. https://en.wikipedia.org/wiki/Arora_%28web_browser%29

Whatever else anyone may come up with.

winhamwr: Makes sense.

src/wymeditor/core.js
```diff
@@ -1140,3 +1140,39 @@ WYMeditor.Helper = {
...
+
+/**
+    WYMeditor._getWymClassForBrowser
+    ================================
+
+    Returns the constructor for the browser-specific wymeditor instance.
+
+    If does not detect a supported browser, returns false;
+*/
+WYMeditor._getWymClassForBrowser = function () {
```
mightyiam: @winhamwr here's your suggested name.

winhamwr: :+1:

src/wymeditor/core.js
```diff
@@ -1140,3 +1140,41 @@ WYMeditor.Helper = {
...
+            return WYMeditor.WymClassBlink;
+        case "safari":
+            return WYMeditor.WymClassSafari;
+    }
+
+    if (jQuery.browser.webkit) {
+        return WYMeditor.WymClassWebKit;
+    }
+
+    WYMeditor.console.warn("WYMeditor could not instantiate: this browser " +
```
mightyiam: Browser not supported warning @winhamwr.

winhamwr:
Looks good!

mightyiam:
Thank you!

PR #665, opened 2015-02-09
Parallel travis without jekyll/docutils requirements

winhamwr:
People should be able to checkout and run the absolute basics without extensive installation.

### README edits

* We don't use IRC
* The documentation is the most important link


winhamwr:
It looks like you published a new version of object-history that completely changed the structure, but you didn't bump the version to mark it as backwards incompatible. Something is also up with combokeys, but I'm not exactly sure what. A manual `npm install combokeys` seems to have fixed that.

winhamwr:
Also needed a `npm install keysim`.

winhamwr:
```
Running "shell:convertReadmeToHomePage" (shell) task
/bin/sh: 1: rst2html: not found
Warning: Command failed: /bin/sh: 1: rst2html: not found
 Use --force to continue.

Aborted due to warnings.
```

The `grunt server` command shouldn't fail because of this. It's pretty mean to anyone just trying to contribute.

winhamwr:
To avoid the failure on rst2html, `sudo apt-get install python-docutils`.

Now I get all of the way to:
```
Running "watch:livereload" (watch) task
Waiting...Waiting...Warning: ENOENT, no such file or directory '/opt/wymeditor/wymeditor/index.html'

Running "watch:jekyll" (watch) task
Waiting...Warning: ENOENT, no such file or directory '/opt/wymeditor/wymeditor/index.html'

Running "watch:jekyll" (watch) task
Waiting...Warning: ENOENT, no such file or directory '/opt/wymeditor/wymeditor/index.html'

Running "watch:jekyll" (watch) task
Waiting...Warning: ENOENT, no such file or directory '/opt/wymeditor/wymeditor/index.html'

```

Which repeats over and over until I get:
```
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.

util.js:35
  var str = String(f).replace(formatRegExp, function(x) {
                      ^
RangeError: Maximum call stack size exceeded
    Warning:  Use --force to continue.

        Aborted due to warnings.

```

winhamwr:
Phew. That Maximum call stack went away after I reloaded my shell.

winhamwr:
I'm not sure of the easiest way to let people without docutils and ruby/jekyll installed still run tests. One way might be something like the [force-task](https://github.com/floriangosse/grunt-force-task) plugin. 

Alternatively, we could go back to supporting the Vagrant development environment so that we can just assume everything works. Supporting Vagrant only makes sense if you're willing to use it for your development, Shahar, otherwise it will get out of date. We need to make sure we have a buildable project, which is not currently the case.

mightyiam:
Very nice, overall. Thank you.

Would you like to address the issues I mentioned yourself, or would you just like to review them and me to fix them?

winhamwr:
Shahar: Any reason that you comment on the commits instead of on the PR itself? Commenting on the PR means I can see all of your comments right in line with the code on one page. With the commits, I can't, and the conversation is lost if it takes multiple future commits to resolve them.

PR #666, opened 2015-02-10
Parser: Auto close tags overcloses TD's in some cases

jlward:


winhamwr:
Jason: You have jshint errors in the build

You need to add QUnit as an external module to the top of the file.
```
$ grunt jshint
Running "jshint:all" (jshint) task
   src/test/unit/specific_feature_tests/xml_parser.js
   2397 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2405 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2414 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2422 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2431 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2439 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2448 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2456 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2465 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2473 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2482 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2490 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2499 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
   2507 |    QUnit.expect(1);
             ^ 'QUnit' is not defined.
>> 14 errors in 25 files
```

winhamwr:
Oops, sorry. It's only for your patch-nested-table-fix branch that the jshint problem exists. I agree with your inaction, there.

jlward:
Yea, jshint is run on travis. So I would be surprised given the green checkbox.

src/wymeditor/parser/xhtml-sax-listener.js
```diff
@@ -172,9 +172,22 @@ WYMeditor.XhtmlSaxListener = function() {
 };
 
 WYMeditor.XhtmlSaxListener.prototype.shouldCloseTagAutomatically = function(tag, now_on_tag, closing) {
-    closing = closing || false;
+    if (tag != 'td' && tag != 'option') {
```
winhamwr: We're only handling autoclosing for these two things because these are two especially problematic tags to leave open. A full solution would work with all of the tags that the HTML5 spec says should autoclose, but we're trying to avoid being a (bad) re-implementation of [parse5](https://github.com/inikulin/parse5). We're doing this because we found a bug in production with nested tables caused by this attempt at a fix.

winhamwr:
> OK this is taking me too much time to really understand.

Sorry about that. We should have added some context via comments on the pull request before asking you to review. I hope it didn't waste too much time. I added inline comments with dddeae8 that should explain the "why" more appropriately. Also [this](https://github.com/wymeditor/wymeditor/pull/666#discussion_r24498649).

I made the 0/undefined change you suggested and clarified the changelog entry to mention the parser. Once the build passes on travis, I'm going to merge this in if I don't hear anything else.

PR #668, opened 2015-02-15
Undo/redo native edits

mightyiam:


src/wymeditor/editor/native-edits.js
```diff
@@ -0,0 +1,38 @@
...
+    WYMeditor.NativeEdits._onNativeEdit
+    ===================================
+
+    Handles native edits for the purpose of Undo/Redo.
+*/
+WYMeditor.NativeEdits.prototype._onNativeEdit = function () {
+    var nativeEdits = this;
+
+    nativeEdits.wym.registerModification();
+};
```
mightyiam: This is pretty much all the code that is required to be added to WYMeditor in order to make use of `edited`.

Modularity FTW.

src/wymeditor/editor/external-modules-browserify.js
```diff
@@ -5,5 +5,6 @@
 // task.
 WYMeditor.EXTERNAL_MODULES = {
     ObjectHistory: require("object-history"),
-    Combokeys: require("combokeys")
+    Combokeys: require("combokeys"),
+    Edited: require("edited")
 };
```
mightyiam: For sake of debugging, this should be split up into one file per external module.

mightyiam: More accurately, this is for the sake of debugging in IE < 11, where there is no source maps support.

src/test/unit/index.html
```diff
@@ -69,6 +69,7 @@
             "specific_feature_tests/class_toggling.js",
             "specific_feature_tests/links.js",
             "specific_feature_tests/dialogs.js",
+            "specific_feature_tests/native-edit-registration.js",
```
winhamwr: All of the other test modules use underscores. Kind of a gross mix, I guess. Maybe something to fix later.

src/test/unit/specific_feature_tests/native-edit-registration.js
```diff
@@ -0,0 +1,55 @@
+/* global
+    QUnit,
+    prepareUnitTestModule,
+    skipKeyboardShortcutTests,
+    simulateKeyCombo
+*/
+
+"use strict";
+
+module("native-edits-registering", {setup: prepareUnitTestModule});
```
winhamwr: The file is `registration` and this module is `registering`. Was that on purpose?

src/test/unit/specific_feature_tests/native-edit-registration.js
```diff
@@ -0,0 +1,55 @@
...
+QUnit.test("edited is instantiated", function () {
+    QUnit.expect(1);
+
+    var wymeditor = jQuery.wymeditors(0),
+        edited = wymeditor.nativeEditRegistration.edited,
+        Edited = WYMeditor.EXTERNAL_MODULES.Edited;
+    QUnit.ok(edited instanceof Edited, "instanceof");
+});
+
+QUnit.test("'edited' seems to be working", function () {
```
winhamwr: How about 'edited' modification callbacks are firing on keystrokes.

That's what this is actually testing.

src/test/unit/specific_feature_tests/native-edit-registration.js
```diff
@@ -0,0 +1,55 @@
...
+    simulateKeyCombo(wymeditor, '3'); // no callback
+    simulateKeyCombo(wymeditor, ' '); // 2nd callback
+    simulateKeyCombo(wymeditor, ' '); // no callback
+    simulateKeyCombo(wymeditor, '.'); // 3rd callback
+
+    callCount = wymeditor.registerModification.callCount;
+    QUnit.strictEqual(callCount, 3, "`registerModification` called 3 times");
+
+    wymeditor.registerModification.restore();
+});
```
winhamwr: I'd really like to see some simple sanity tests with different combinations of events followed by undo/redo and tests on the new editor content. Keystrokes, toolbar clicks, container type changes, insertions via dialog, class changes, etc.

We don't need full test coverage on corner cases, but we need to know that mixing keystrokes with non-native things has sane undo/redo behavior. Does that seem like a good testing strategy to you?

winhamwr:
What should be working, as a result of this PR?

src/test/unit/specific_feature_tests/native-edit-registration.js
mightyiam: As far as unit tests go, I'm happy with these.

I will add some external test. First, to Edited and then to WYMeditor.

Then, we'll be able to say for certain that our code works.

mightyiam: We can't test content changes in this regard using unit tests because we are only simulating keyboard events. The native actions do not really occur. So there is no content change. Attempting to simulate the content changes will be complex and confusing.

External tests are what we're after, here.

I don't yet start that, so I think we can quite safely merge this in. These tests are enough because we're testing that Edited is operational.

We're trusting that it is doing its part.

It is far better to have further confirmation, in the general `undo_redo.js` tests file, that native and editor modifications are getting undone/redone as expected. But, again, that should be done using external test.

mightyiam:
Naming issues fixed.

Please see my buried up comments.

As a result of this PR it is possible to undo and redo native edits that are listed [here](https://github.com/mightyiam/edited#how-does-it-work).

I'm aware of some issues but this is better than nothing.

winhamwr:
I'm worried about the testing on this, especially the behavior for mixing native and non-native events, but we'll call it progress.

mightyiam:
Awesome!

PR #670, opened 2015-02-18
focusDocumentBody

mightyiam:


src/wymeditor/editor/base.js
```diff
@@ -278,16 +278,16 @@ WYMeditor.editor.prototype.get$Buttons = function () {
...
     Sets focus on the document.
 */
-WYMeditor.editor.prototype.focusOnDocument = function () {
+WYMeditor.editor.prototype.focusDocumentBody = function () {
     var wym = this,
-        doc = wym._iframe.contentWindow;
+        body = wym.body();
 
-    doc.focus();
+    body.focus();
```
mightyiam: This seems to be a better approach.

winhamwr: What behavior does this correct? Why is it a better approach?

PR #671, opened 2015-02-21
browserified modules: split them up

mightyiam:


mightyiam:
Bundles produced by Browserify are single files which contain the code of the desired module (point of entry) and the codes of its entire dependency tree (derived from `require` calls).

Therefore, these files can get quite large and this makes debugging hard or even impossible (read: some IEs).

So, source maps to the rescue. They are produced by Browserify (we have that configured in the grunt task by `debug: true`).

But IE has source maps support since only IE11 (I haven't yet tried [this](http://vswebessentials.com/features/javascript#source-maps) yet)!

So the least we can do is make individual bundles.

That is what this PR does.

winhamwr:
> So the least we can do is make individual bundles.

:+1:

PR #672, opened 2015-02-22
Save state before undo (if no redos exist)

mightyiam:


mightyiam:
This fixes an issue.

After an undo, redo is possible up to the last saved state.

This makes sure that the last saved state is from when you last hit undo.

If the thing you have just done is hit undo and then you hit undo again then that's a case where we do not wish to save state.

winhamwr:
The lack of tests is scary, here.

winhamwr:
Can't we write a test for this using toolbar actions?

src/wymeditor/editor/undo-redo.js
```diff
@@ -79,6 +79,7 @@ WYMeditor.UndoRedo.prototype._do = function (what) {
         if (history.changesetsBack.length === 0) {
             return;
         }
+        if (!history.changesetsFore.length) undoRedo._add()
```
winhamwr: Also, the fact that we need to use an underscore method makes me think that either:

* This should be handled in `history` via a public method
* `_add()` should be `add()`

if the only existing implementation of a library requires usage of private methods, that's a good sign we can improve the API.

src/wymeditor/editor/undo-redo.js
```diff
@@ -79,6 +79,9 @@ WYMeditor.UndoRedo.prototype._do = function (what) {
         if (history.changesetsBack.length === 0) {
             return;
         }
+        if (!history.changesetsFore.length) {
+            undoRedo._add()
```
mightyiam: Style fixed.

> Also, the fact that we need to use an underscore method makes me think that either:

> * This should be handled in `history` via a public method
> * `_add()` should be `add()`

> if the only existing implementation of a library requires usage of private methods, that's a good sign we can improve the API.

Trying to understand what you're saying here it comes to my mind that perhaps you thought that `_add` is a method of an instance of object-history—that `undoRedo` here is an instance of object-history.

Is that true?

Otherwise, I didn't get where you're coming from.

mightyiam:
> The lack of tests is scary, here.

I'm confident.

> Can't we write a test for this using toolbar actions?

This really really calls for external tests. Which we should write pronto.

I suggest we

1. merge this, #673 and #674 
1. cook a build for limited distribution in PS
1. https://trello.com/c/NONYrxxY/16-edited-should-not-callback-on-the-very-first-edit
1. https://trello.com/c/YviOO3pz/10-external-tests-for-edited
1. https://trello.com/c/lg2287nA/18-external-tests-for-wymeditor-undo-redo
1. Fix any reported regression
1. Release!

src/wymeditor/editor/undo-redo.js
winhamwr: Yup. That was my mistake, exactly. Sorry about that.

This still needs a semicolon, though.

winhamwr:
I'm not sure why Github isn't showing the Travis build status, but this change caused [failing tests](https://travis-ci.org/wymeditor/wymeditor/jobs/51986993). Once those are working, unless you'd like further review, I'm +1 on merging.

> I suggest we

I like this plan. I'll cut a 1.0.0-rc.1.patch3 release and we'll try it internally at PolicyStat. Then, step 7 is version 1.0.0

mightyiam:
Depends on [this](https://trello.com/c/VWY6Hdyl/20-edited-anyedit-callback).

winhamwr:
The card this depends on was archived. Does that mean this is ready?

mightyiam:
> The card this depends on was archived. Does that mean this is ready?

I apologize. I'm in a process of moving to waffle and I'm doing it a bit messy.

It still depends on https://github.com/PolicyStat/edited/issues/29.

mightyiam:
Dependency met. Continuing.

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -19,20 +19,36 @@ WYMeditor.NativeEditRegistration = function (wym) {
     // https:/github.com/mightyiam/edited
     nativeEditRegistration.edited = new WYMeditor.EXTERNAL_MODULES.Edited(
         wym._doc.body,
-        function () {
-            nativeEditRegistration._onNativeEdit.call(nativeEditRegistration);
-        }
+        nativeEditRegistration._onSensibleNativeEdit
+            .bind(nativeEditRegistration),
+        nativeEditRegistration._onAnyNativeEdit
+            .bind(nativeEditRegistration)
```
mightyiam: ES5 `Function.prototype.bind`.
https://trello.com/c/QRHvjTH9/1115-depend-on-require-include-es5-shim

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -19,20 +19,36 @@ WYMeditor.NativeEditRegistration = function (wym) {
     // https:/github.com/mightyiam/edited
     nativeEditRegistration.edited = new WYMeditor.EXTERNAL_MODULES.Edited(
         wym._doc.body,
-        function () {
-            nativeEditRegistration._onNativeEdit.call(nativeEditRegistration);
```
mightyiam: BTW this line has a mistake. The call to `Function.prototype.call` is unnecessary. It could have been just `nativeEditRegistration._onNativeEdit()`.

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -19,20 +19,36 @@ WYMeditor.NativeEditRegistration = function (wym) {
...
+        nativeEditRegistration._onAnyNativeEdit
+            .bind(nativeEditRegistration)
     );
 };
 
 /**
-    WYMeditor.NativeEditRegistration._onNativeEdit
-    ==============================================
+    WYMeditor.NativeEditRegistration._onSensibleNativeEdit
+    ======================================================
```
mightyiam: Now we give two callbacks to Edited instantiation: 'sensible' and 'any'.

package.json
```diff
@@ -36,7 +36,7 @@
   ],
   "devDependencies": {
     "combokeys": "^2.0.0",
-    "edited": "^1.0.6",
+    "edited": "^1.1.1",
```
mightyiam: Has https://github.com/PolicyStat/edited/pull/37.

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -19,20 +19,40 @@ WYMeditor.NativeEditRegistration = function (wym) {
...
+
+    Handles all native edits.
+*/
+WYMeditor.NativeEditRegistration.prototype._onAnyNativeEdit = function () {
+    var nativeEditRegistration = this,
+        undoRedo = nativeEditRegistration.wym.undoRedo;
+
+    // remove redo points
+    undoRedo.history.changesetsFore = [];
+    // If a user is in the middle of typing their first word, there are changes,
```
winhamwr: Made this a useful "why" comment because I'm fond of future us.

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -19,20 +19,41 @@ WYMeditor.NativeEditRegistration = function (wym) {
...
+*/
+WYMeditor.NativeEditRegistration.prototype._onAnyNativeEdit = function () {
+    var nativeEditRegistration = this,
+        undoRedo = nativeEditRegistration.wym.undoRedo;
+
+    // remove redo points
+    undoRedo.history.changesetsFore = [];
+    // this is a flag for future actions. For example, on undo, when this flag
+    // is true, the state is saved as an undo point before undoing. Otherwise,
+    // this state would have been lost forever.
```
mightyiam: @winhamwr Is this good?

winhamwr: No. This explains what it does, not why "this state would have been lost forever". The reason this code needs to exists is because of the difference between sensible edits and any edits, which is an abstraction that can't be assumed for the person reading this code (future us).

The "on undo, when this flag is true, the state is saved as an undo point before undoing." is a bad comment because it's not near the code it's referring to and it's just an english description of some code's behavior (a "what" comment). So it doesn't actually help you understand the code and is likely to lie in the future.

src/wymeditor/editor/undo-redo.js
```diff
@@ -111,6 +112,10 @@ WYMeditor.UndoRedo.prototype._do = function (what) {
         if (history.changesetsBack.length === 0) {
             return;
         }
+        if (undoRedo.hasUnregisteredEdits) {
+            // in order to be able to 'redo' to this yet unregistered state
```
winhamwr: "unregistered" is not a useful term, here. It's ambiguous. Registered where? Why would it not be registered?

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -19,20 +19,49 @@ WYMeditor.NativeEditRegistration = function (wym) {
...
+    // Non-native modifications are registered when they are performed.
+    // Contrary to those, only a part of the native edits are
+    // registered when they are performed.
+    // A state with unregistered modifications will be lost upon an undo
+    // and redo because it was never registered.
+    // The last registered state doesn't include it and that will be the result
+    // of the undo and redo operation.
+    // This flag lets the undo operation know that it should add the current
+    // state as a history point (register it) before undoing.
+    // Then the next redo will revert to this state.
```
mightyiam: @winhamwr I'm no big fan of the "why" word. I'm a fan of a wide understanding.

To me, most human answers to "why" are narrow points of view.

Trying to answer "why" is usually by asking what could be expected to have been different that would have prevented the outcome that we're asking about.

That's guess work.

I tried to provide just enough background here for a sufficient understanding.

Does it work for you?

src/wymeditor/editor/undo-redo.js
```diff
@@ -88,6 +88,7 @@ WYMeditor.UndoRedo.prototype._add = function () {
         wym = undoRedo.wym;
 
     undoRedo.history.add(wym.getCurrentState());
+    undoRedo.hasUnregisteredModification = false;
```
mightyiam: @winhamwr I changed the name from `hasUnregisteredEdits` to `hasUnregisteredModification`.

This name is more suitable because of the existing `registerModification` method.

src/wymeditor/editor/native-edit-registration.js
winhamwr: You don't have to guess at the "why". You just have to explain it and attempt to simulate the state of yourself after not seeing this code for 12 months. Some folks have trouble with that, which is where code review comes in.

This is a pretty good comment block, but the thing that future-us won't understand from reading this is that native edits are only registered on effectively word/sentence boundaries.

My edit:

Contrary to those, native edits (e.g. typing) aren't registered until word boundaries are reached (space, enter) or edit actions (backspace, delete, cut, paste, etc). This handles an undo after e.g. only part of word is typed.

winhamwr: Made my attempt at adding details re: unregistered edits.

winhamwr:
Looks good. If you're cool with the gist of my edit, I think this is ready to merge.

src/wymeditor/editor/native-edit-registration.js
mightyiam: "This handles an undo after e.g. only part of word is typed." is unclear to me.

mightyiam: I wouldn't want to have the last say in this.

mightyiam:
:open_mouth: exciting.

Issue #685, opened 2015-03-03
Some recent parser tests fail in IE8


mightyiam:
#666

winhamwr:
Doh. Did we not test these in IE8 @jlward ?

mightyiam:
In any case, these seem like pretty much meaningless attempts from IE8 to make our tests fail. An extra space here and there.

mightyiam:
On it.

winhamwr:
Let's just make the expected output different for IE8. It's not worth doing actual work to strip newlines in the parser unless the newlines alter some semantic meaning (like with the bogus trailing newlines in `li` elements inserted by IE).

mightyiam:
This is actual output from IE8. It seems to provide the broken, unclosed tag HTML back at us:
![screenshot from 2015-03-06 17 44 52](https://cloud.githubusercontent.com/assets/635591/6528127/a07e4e6a-c428-11e4-8c1e-13bff02dbefd.png)

Besides being utterly strange to me (what kind of a design causes this?) it also has a quirk of effectively adding a newline after the AAA, which gets turned into a space at some point before the assertion.

What use does the newline in `<p>hello\nshahar</p>` have?

## Currently in WYMeditor in chromium:
![screenshot from 2015-03-06 19 57 38](https://cloud.githubusercontent.com/assets/635591/6530847/10e14cb6-c43d-11e4-98f0-7e1e395859cf.png)

## In IE8
![screenshot from 2015-03-06 20 17 46](https://cloud.githubusercontent.com/assets/635591/6530965/ecb2541a-c43d-11e4-8995-51143438afa3.png)

The two look the same in the editor but the output is different. Chrome outputs a newline and IE8 a space.

So newlines anywhere but in `pre` are replaced by spaces in the DOM but may be output as either newlines or spaces when requesting HTML.

A consistent policy should be declared on newlines, turning them into spaces by the parser.

The case of the missing tag in IE should be fixed in some way, ideally. But it may as well just be accepted, as you say.

mightyiam:
Anyway, I'm unmarking as a regression because the failing tests are very minor.

winhamwr:
![screenshot from 2015-03-06 17 44 52](https://cloud.githubusercontent.com/assets/635591/6528127/a07e4e6a-c428-11e4-8c1e-13bff02dbefd.png)

Crap. Does that mean that #666 broke the TD auto-closing behavior in IE8?

mightyiam:
[...]

PR #688, opened 2015-03-06
Parser: handle newlines that are not in 'pre'

mightyiam:
Fix tests for this

### DIFF
--- a/src/test/unit/specific_feature_tests/lists.js
+++ b/src/test/unit/specific_feature_tests/lists.js
@@ -2812,24 +2812,20 @@ test("Tab has no effect outside lists", function () {
 
 module("list-newline_normalization", {setup: prepareUnitTestModule});
 
-test("Shouldn't eat newline text spacing in li", function () {
+test("Should eat newline text spacing in li", function () {
     QUnit.expect(1);
 
     var initHtml = String() +
             '<ul>' +
                 '<li>' +
-                    'Lorem ipsum dolor' + "\n" +
-                    'sit amet, consectetur' + "\n\n" +
-                    'adipiscing elit. Integer' + "\n\n\n" +
-                    'sagittis porta dapibus.' +
+                    'Lorem ipsum dol\nor' +
                 '</li>' +
             '</ul>',
 
         expectedHtml = String() +
             '<ul>' +
                 '<li>' +
-                    'Lorem ipsum dolor sit amet, consectetur adipiscing elit.' +
-                    ' Integer sagittis porta dapibus.' +
+                    'Lorem ipsum dolor' +
                 '</li>' +
             '</ul>',
         wymeditor = jQuery.wymeditors(0);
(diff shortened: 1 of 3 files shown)


src/test/unit/specific_feature_tests/lists.js
```diff
@@ -2812,24 +2812,20 @@ test("Tab has no effect outside lists", function () {
...
+                    'Lorem ipsum dol\nor' +
                 '</li>' +
             '</ul>',
 
         expectedHtml = String() +
             '<ul>' +
                 '<li>' +
-                    'Lorem ipsum dolor sit amet, consectetur adipiscing elit.' +
-                    ' Integer sagittis porta dapibus.' +
+                    'Lorem ipsum dolor' +
```
mightyiam: Not only in lists, but everywhere that is not a `pre`, actually.

mightyiam:
What I'm concerned about in this approach is whether some paste from word things work as expected or not. Perhaps Word provides some linebreaks which actually should be interpreted as `br` tags?

mightyiam:
If we choose to proceed with this change, then I would write two more tests:

- [ ] newlines are stripped in a few different elements
- [ ] newlines are kept in `pre` elements

src/wymeditor/parser/xhtml-sax-listener.js
```diff
@@ -284,13 +284,10 @@ WYMeditor.XhtmlSaxListener.prototype.getTagForStyle = function (style) {
...
-        // We should strip trailing newlines from text inside li tags because
-        // IE adds random significant newlines inside nested lists
-        text = text.replace(/(\r|\n|\r\n)+$/g, '');
-
-        // Let's also normalize multiple newlines down to a single space
-        text = text.replace(/(\r|\n|\r\n)+/g, ' ');
+    if (!(this.last_tag && this.last_tag === 'pre')) {
+        // We respect newlines only when in 'pre'. Outside of 'pre',
+        // we remove all of them.
+        text = text.replace(/(\r|\n|\r\n)/g, '');
```
winhamwr: We can't just remove them. A single newline is significant.

These two things are actually different:

`<p>hello\nshahar</p>`

`<p>helloshahar</p>`

mightyiam:
The last commit introduces a solution that seems excellent to me. It also fixes the failing tests from #685.

From the commit:
``` js
if (!(this.last_tag && this.last_tag === 'pre')) {
    // We respect newlines as they are, only when in 'pre'
    // outside of 'pre',
    // at end of elements they are removed
    text = text.replace(/(\r|\n|\r\n)+$/g, '');
    // the rest are replaced with spaces
    text = text.replace(/(\r|\n|\r\n)/g, ' ');
}
```

And here is what it looks like:

## Chrome
screenshot from 2015-03-06 20 38 21

## IE8
screenshot from 2015-03-06 20 39 26

If you OK this I'll write those two tests.

mightyiam:
Please see this behavior as an improvement over https://github.com/wymeditor/wymeditor/issues/685#issuecomment-77609999.

winhamwr:
I knew I felt weird about this and I finally figured out why. The way that whitespace is handled within an element is actually defined in the [html5 spec](http://www.w3.org/TR/CSS2/text.html#white-space-model). The spec makes it clear that the rendered behavior of whitespace can be controlled with the [white-space](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) css property.

That means that, at the HTML-parsing level, we can't safely remove whitespace arbitrarily. For example, it wouldn't be rare to use a `code {white-space: pre}` CSS declaration to make rendering code samples whitespace-sensitive.

PR #689, opened 2015-03-06
Do not initialize on IE < 8

mightyiam:


mightyiam:
@winhamwr, please review.

PR #697, opened 2015-03-08
Travis: use default node.js version

mightyiam:


.travis.yml
```diff
@@ -1,6 +1,4 @@
 language: node_js
-node_js:
-    - 0.10
```
mightyiam: This is a number. That seems likely the reason that, we end up testing against Node.JS `0.1`.

We could use a string, but I've good experience with leaving the default.

If/when it fails, we can either fix it or pin the version.

mightyiam:
It works.

mightyiam:
Please review.

mightyiam:
It [only seems] as if it is using Node.js "0.1". Click on a build to see that it actually used `0.10`.

Still, I like this change.

winhamwr:
:+1:

mightyiam:
Thank you.

PR #701, opened 2015-03-23
Register container changes

mightyiam:


src/test/unit/specific_feature_tests/set-root-container.js
```diff
@@ -0,0 +1,136 @@
...
+        // h1, h2, etc.
+        loadDefaultPlugins: false
+    });
+}});
+
+// We test that each tag (as in element) can be changed to each tag.
+// This creates this array of tags.
+var pairs = [];
+// what about the div?
+['p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'pre']
```
mightyiam: What about when the default root container is a `div`?

mightyiam:
The tests for `setRootContainer` may be incomplete but they're more than the zero tests we had until now. Please review.

src/test/unit/utils.js
```diff
@@ -134,10 +140,12 @@ function prepareUnitTestModule(args) {
     }
 
     args.options.postInit = function (wym) {
-        // TODO: We should not load all these plugins by default.
-        wym.listPlugin = new ListPlugin({}, wym);
-        wym.tableEditor = wym.table();
-        wym.structuredHeadings();
+        if (args.loadDefaultPlugins) {
+            // TODO: We should not load all these plugins by default.
```
winhamwr: I'd say that the listPlugin and tableEditor should be loaded in >90% of use cases and we need to make sure they're test. In a perfect world, we would have tests both with them and without them, but that seems like a lot of extra tests.

`structuredHeadings` is a bit of a special case, at least right now.

winhamwr:
> incomplete but they're more than the zero tests we had until now

Agreed.

mightyiam:
Awesome! Thank you!

PR #709, opened 2015-04-19
Undoing back to zero loses selection

mightyiam:
Instantiate object-history on body focus

### DIFF
--- a/src/wymeditor/editor/undo-redo.js
+++ b/src/wymeditor/editor/undo-redo.js
@@ -48,11 +48,12 @@ WYMeditor.UndoRedo = function (wym) {
 
     It makes sure that the first undo point contains a selection.
 
-    The first selection is made along with the first focus.
+    After the editor's instantiation there is no selection. The first
+    selection is made along with the first focus.
 
     This method will run before the native action (in which the selection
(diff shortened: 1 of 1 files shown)


src/wymeditor/editor/undo-redo.js
```diff
@@ -22,10 +22,6 @@ WYMeditor.UndoRedo = function (wym) {
 
     undoRedo.wym = wym;
 
-    // https://github.com/mightyiam/object-history
-    undoRedo.history = new WYMeditor.EXTERNAL_MODULES
-        .ObjectHistory(wym.getCurrentState());
```
mightyiam: This always produces a first undo point that is lacking selection.

That means that when undoing all the way back to the first point makes the selection go away, meaning the user must place the caret in order to type characters again.

mightyiam:
In tests that test undo/redo, a [`focusOnDocument`](https://wymeditor.readthedocs.org/en/latest/writing_plugins/api.html#focusondocument) can help.

In `manipulationTestHelper` a `wym.undoRedo.reset` call results in a (re)initialization of object-history.

mightyiam:
Thank you!

PR #710, opened 2015-04-23
Seamless to default

winhamwr:
* Make the `seamless` skin the default skin. It has much better usability.
* Let skins define their own `iframeHtml` and use that preferentially over anything not directly overridden. This allows us to maintain backwards compatibility.


winhamwr:
@mightyiam I think the seamless skin is the way to go with 1.0 and this iframeHtml trick keeps backwards compatibility. Thoughts?

winhamwr:
We should namespace all the attributes that can be overridden by skins.

src/wymeditor/core.js
```diff
@@ -775,17 +775,30 @@ jQuery.fn.wymeditor = function (options) {
...
         preBind: null,
         postInit: null
+    }
+
+    mergedOptions = jQuery.extend(defaultOptions, options);
 
-    }, options);
+    // If an iframeHtml option wasn't included, default to using the iframeHtml
+    // defined by the skin.
+    if (typeof options.iframeHtml !== "undefined") {
```
mightyiam: I would write
```js
if (!options.iframeHtml) {
}
```
It is a bit easier on the eyes, for me. It also resolves to `true` when `options.iframeHtml === ''` (an empty string).
screenshot_2015-04-24_13-02-06

mightyiam: Oh, oops. That is the reverse logic:
screenshot_2015-04-24_13-04-46

So I would write
```js
if (options.iframeHtml) {
// ...
}
```

src/wymeditor/core.js
```diff
@@ -775,17 +775,30 @@ jQuery.fn.wymeditor = function (options) {
...
+        // If the user passed an iframeHtml, use that
+        if (typeof WYMeditor.SKINS[mergedOptions.skin] !== "undefined") {
```
mightyiam: Same as above, I would write
```js
if (WYMeditor.SKINS[mergedOptions.skin]) {
}
```
screenshot_2015-04-24_13-26-04

mightyiam:
But I don't completely understand this.

* What is is `iframeHtml` option? It doesn't seem to me like something that I can easily document. It doesn't seem to me like something I would like to make public API. I don't remember using it myself.

The seamless skin seems to use a polling technique to resize the document iframe. To me this seems very hackish. Whenever I make the document grow bigger in height (by creating a paragraph) I can see the iframe's scrollbar appearing and disappearing again. That is not the impression that I would like to make. If "seamlessness" can be achieved without the scrollbar appearing at all (preferably based on events—not polling) then it would seem to me as an improvement.

winhamwr:
> What is is iframeHtml option?

It's one of the already-documented [template options](https://wymeditor.readthedocs.org/en/v1.0.0-rc.2/customizing_wymeditor/index.html#list-of-html-template-options).

The [seamless example](https://github.com/wymeditor/wymeditor/blob/master/src/examples/20-seamless.html#L68) is one place where we had an example using it.

> The seamless skin seems to use a polling technique to resize the document iframe.

A pure event-based approached isn't feasible. In order to calculate the newly-desired size, you must access several `height()` calls. Making those calls requires the browser to do a full DOM layout recalculation, which is basically the most expensive thing you can make the browser do. There is an optimization to be made for modern browsers by combining events + rate limiting (polling) + [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame), but it's definitely a polish item.

src/wymeditor/core.js
winhamwr: :+1:

src/wymeditor/core.js
```diff
@@ -527,6 +584,239 @@ jQuery.extend(WYMeditor, {
     EXTERNAL_MODULES : {}
 });
 
+WYMeditor.DEFAULT_OPTIONS = {
```
winhamwr: I moved this to an externally-referenceable constant so that it can be used in tests.

src/wymeditor/core.js
```diff
@@ -527,6 +584,239 @@ jQuery.extend(WYMeditor, {
...
+    // These values will override the default for their respective
+    // `DocumentStructureManager.structureRules`
+    structureRules: {
+        defaultRootContainer:  'p'
+    },
+
+    boxHtml: String() +
+        "<div class='wym_box'>" +
+            "<div class='wym_area_top'>" +
+                WYMeditor.TOOLS +
```
winhamwr: Because it includes these variable references, this can't be declared with the rest of the `WYMeditor` attributes.

src/wymeditor/core.js
```diff
@@ -541,251 +831,39 @@ jQuery.extend(WYMeditor, {
...
-        preInit: null,
-        preBind: null,
-        postInit: null
-
-    }, options);
-
-    options = jQuery.extend(WYMeditor.DEFAULT_DIALOG_OPTIONS, options);
+
+    mergedOptions = jQuery.extend(optionsWithDefaults, optionsWithSkinOverrides);
+    mergedOptions = jQuery.extend({}, WYMeditor.DEFAULT_DIALOG_OPTIONS, mergedOptions);
```
winhamwr: Added the `{}` here because we were previously actually modify the `DEFAULT_DIALOG_OPTIONS` in place, which was a bug.

src/test/unit/specific_feature_tests/skins.js
```diff
@@ -0,0 +1,36 @@
...
+);
+
+test("skin's iframeHtml is used preferentially", function () {
+    var wym = jQuery.wymeditors(0);
+    QUnit.expect(2);
+
+    // Ensure that the basic skin has a non-default iframeHtml, so that this
+    // test is meaningful
+    notEqual(
+        WYMeditor.DEFAULT_OPTIONS.iframeHtml,
```
winhamwr: Added this check to protect against future changes to `DEFAULT_OPTIONS` or the basic skin possibly making this test not actually test what it needs to.

src/test/unit/test.js
```diff
@@ -59,6 +60,23 @@ test("Instantiate", function () {
               "Type of first WYMeditor instance, using jQuery.wymeditors(0)");
 });
 
+asyncTest("Custom dialog options don't change future defaults", function () {
```
winhamwr: This tests the bug with `DEFAULT_DIALOG_OPTIONS` in-place modification. It failed until I added the `{}`.

winhamwr:
I still need to do a changelog and documentation updates for this, but I think the actual modification + tests is ready for review.

CHANGELOG.md
```diff
@@ -47,6 +47,9 @@ WYMeditor.
     but it is recommended that existing custom dialogs be rewritten using the
     new `wym.dialog` API (see API documentation) or at least that they be
     tested.
+  * All toolbar buttons that open a dialog must include the `wym_opens_dialog`
+    class or they'll be opened in the background by a new, more-consistent
+    method of returning focus to the editor after button clicks.
```
mightyiam: I must have missed this.

I feel that we can gain a lot more clarity in this sentence with a small restructure.

Something to do with the separation of its two parts (as connected by "by").

CHANGELOG.md
```diff
@@ -235,7 +238,7 @@ All of these changes are documented in detail in the
   you're using the default skin, you'll need to add a line like the
   following:
   `<link rel="stylesheet" type="text/css"
-    href="wymeditor/skins/default/skin.css" />`
+    href="wymeditor/skins/basic/skin.css" />`
```
mightyiam: Does this change need to be documented in any other part of the changelog?

What about upgrades to integrations that currently use the seamless skin?

What about mentioning that the default skin changed, in general?

src/test/unit/specific_feature_tests/skins.js
```diff
@@ -0,0 +1,36 @@
...
+*/
+"use strict";
+
+module(
+    "skin-ui-overrides",
+    {
+        setup: function () {
+            prepareUnitTestModule({
+                options: {skin: 'basic'}
+            });
```
mightyiam: Yey! @winhamwr used my helper.

src/test/unit/specific_feature_tests/skins.js
```diff
@@ -0,0 +1,36 @@
...
+
+    // Ensure that the basic skin has a non-default iframeHtml, so that this
+    // test is meaningful
+    notEqual(
+        WYMeditor.DEFAULT_OPTIONS.iframeHtml,
+        WYMeditor.SKINS.basic.UI_COMPONENTS.iframeHtml
+    );
+
+    // Ensure that non-default attribute propagates
+    equal(wym._options.iframeHtml, WYMeditor.SKINS.basic.UI_COMPONENTS.iframeHtml);
```
mightyiam: `strictEqual`, ideally. Just as use of `==` is not recommended.

src/test/unit/test.js
```diff
@@ -59,6 +60,23 @@ test("Instantiate", function () {
...
+    var startingDefault = WYMeditor.DEFAULT_DIALOG_OPTIONS.hrefSelector;
+    QUnit.expect(3);
+
+    notEqual(startingDefault, '.wym_customized_thing');
+
+    vanishAllWyms();
```
mightyiam: @winhamwr used my helper again! :laughing:

src/wymeditor/core.js
```diff
@@ -474,6 +474,63 @@ jQuery.extend(WYMeditor, {
         TEXT: 3
     },
 
+    // These UI components will preferentially use those defined in the chosen
+    // skin inside `UI_COMPONENTS` instead of the defaults.
```
mightyiam: I feel that a big readability improvement can be easily gained by replacing "those" with a something more descriptive.

src/wymeditor/core.js
```diff
@@ -474,6 +474,63 @@ jQuery.extend(WYMeditor, {
...
+        'dialogLinkHtml',
+        'dialogFeatures',
+        'dialogImageHtml',
+        'dialogTablehtml',
+        'dialogPastehtml',
+        'dialogPreviewHtml'
+    ],
+
+    // These UI component selectors will preferentially use those defined in
+    // the chosen skin inside `UI_COMPONENT_SELECTORS` instead of the defaults.
```
mightyiam: Same as above.

src/wymeditor/core.js
mightyiam: Seems wise.

src/wymeditor/core.js
```diff
@@ -541,251 +831,39 @@ jQuery.extend(WYMeditor, {
...
-        updateEvent:       "click",
-
-        stringDelimiterLeft:  "{",
-        stringDelimiterRight: "}",
-
-        preInit: null,
-        preBind: null,
-        postInit: null
-
-    }, options);
```
mightyiam: It is nice to have this out of here.

If we were using a module system, it would have had its own file.

src/wymeditor/core.js
mightyiam: I don't get the point.

For starters, where is "the rest of the `WYMeditor` attributes?

src/wymeditor/core.js
```diff
@@ -541,251 +831,39 @@ jQuery.extend(WYMeditor, {
...
-        },
+jQuery.fn.wymeditor = function (providedOptions) {
+    var $textareas = this,
+        mergedOptions,
+        optionsWithDefaults,
+        optionsWithSkinOverrides,
+        skin;
+
+
+    optionsWithDefaults = jQuery.extend({}, WYMeditor.DEFAULT_OPTIONS, providedOptions);
```
mightyiam: Good catch.

src/test/unit/specific_feature_tests/skins.js
mightyiam: This is an example of what I asked you about. It is so easy to make this mistake and produce this bug.

src/wymeditor/core.js
mightyiam: We do have some kind of multi–file system. How about pulling this out into its own file?

src/wymeditor/core.js
mightyiam: Good catch.

src/wymeditor/core.js
mightyiam: We have four `jQuery.extend`s and:

* `optionsWithDefaults`
* `optionsWithSkinOverrides`
* `mergedOptions`

And `mergedOptions` is merged into twice.

Consistency would look better, in my taste.

How about we drop `optionsWithDefaults` and `optionsWithSkinOverrides`?

mightyiam: And how about [merging them all in one call](https://api.jquery.com/jQuery.extend/)?

src/wymeditor/core.js
```diff
@@ -944,6 +1022,44 @@ WYMeditor._computeBasePath = function (wymPath) {
     return wymPath.substr(0, lastSlashIndex + 1);
 };
 
+WYMeditor._getSkinUIOverrides = function (providedOptions, skin) {
```
mightyiam: Some documentation would be useful for me to understand.

Also, I feel that this static method may be easier to understand if it also did the merging and not only provide the object for the merge.

mightyiam:
Regarding the feasibility of an event–based means of automatic readjusting of the height,

Do you mean [`jQuery.height`](https://api.jquery.com/height/) calls? Are those very expensive?

Even if they were expensive, would we care? It happens only when certain events are fired, I guess.

I don't think I should put my time into this in development—only in some understanding.

winhamwr:
> Do you mean jQuery.height calls? Are those very expensive?

Yes, and yes.

> Even if they were expensive, would we care? It happens only when certain events are fired, I guess.

We would need to fire them on basically every keystroke, because each keystroke has the potential to wrap a line, requiring more height.

CHANGELOG.md
winhamwr: See [this comment](https://github.com/wymeditor/wymeditor/pull/710#issuecomment-96103136).

mightyiam:
So you're ok with the flashing scrollbar?

winhamwr:
> So you're ok with the flashing scrollbar?

I definitely don't love it. I greatly prefer it to the constant double scrollbar in the old default skin and the usability headaches that it caused. Part of my impetus for getting this skin in as the default is to make sure that any future improvements benefit both the FOSS users and PolicyStat users.

mightyiam:
@winhamwr this seems to be the last open issue for the `v1.0.0` milestone.

How do you suggest we continue?

src/wymeditor/core.js
```diff
@@ -944,6 +1022,44 @@ WYMeditor._computeBasePath = function (wymPath) {
...
+    var uiOverrides = {},
+        skinOverride;
+
+    if (!providedOptions) {
+        return uiOverrides;
+    }
+
+    if (skin.UI_COMPONENTS) {
+        jQuery.each(WYMeditor.UI_COMPONENTS, function (i, componentAttr) {
```
mightyiam: We [can rely on having ES5](https://github.com/wymeditor/wymeditor#requirements) and the iterable is an array so this should be using `Array.prototype.forEach`, instead of `jQuery.each`.

src/wymeditor/core.js
```diff
@@ -944,6 +1022,44 @@ WYMeditor._computeBasePath = function (wymPath) {
...
+                skinOverride = skin.UI_COMPONENTS[componentAttr];
+                if (typeof skinOverride !== "undefined") {
+                    // Skins can use empty strings for overrides
+                    uiOverrides[componentAttr] = skinOverride;
+                }
+            }
+        });
+    }
+    if (skin.UI_COMPONENT_SELECTORS) {
+        jQuery.each(WYMeditor.UI_COMPONENT_SELECTORS, function (i, selectorAttr) {
```
mightyiam: `Array.prototype.forEach`.

mightyiam:
FYI I've just merged `master` into this.

mightyiam:
Bang. Is this expected to work with all IE versions that we support?

Testing now, the only IE version in which the height of the Iframe is growing, is IE11.

Firefox and Chrome work.

IE10 and below have the scroll bar and the height of the Iframe is not increased.

I suggest we release `v1.0.0` without this.

mightyiam:
Any thoughts?

mightyiam:
But, otherwise, in IE10 and below, the skin looks fine and the editor works while the height of the Iframe remains static and unchangeable.

Unchangeable seems to be the case in `master`, as well, so perhaps in IE10 and below, the lack of growing Iframe height can be considered a graceful degradation.

More points for graceful degradation is that in IE10 and below there aren't any warnings/errors in the console when the height doesn't change.

mightyiam:
Actually, only now I've pushed the merging of `master` in.

CHANGELOG.md
mightyiam: May I suggest the following, instead?
```
* All toolbar buttons that open a dialog must have the `wym_opens_dialog` class.
  Otherwise, the dialog window might end up in the background, hidden from the user`
```

src/test/unit/specific_feature_tests/skins.js
```diff
@@ -0,0 +1,36 @@
...
+    prepareUnitTestModule
+*/
+"use strict";
+
+module(
+    "skin-ui-overrides",
+    {
+        setup: function () {
+            prepareUnitTestModule({
+                options: {skin: 'basic'}
```
mightyiam: Is this option, required? Isn't `'basic'` the default now?

mightyiam: According to [this line in the test below](https://github.com/wymeditor/wymeditor/pull/710/files#diff-6567671bc27f9cb5c72677b1981b278aR35), it seems that `'basic'` is not the default.

I'm confused, because [changes like this](https://github.com/wymeditor/wymeditor/pull/710/files#diff-209fc1960195c91ceff0368bcbe4b835L9) make it seem that we are changing the default to be `'basic'`.

src/test/unit/test.js
```diff
@@ -59,6 +60,23 @@ test("Instantiate", function () {
...
+    jQuery('.wym').wymeditor({
+        hrefSelector: '.wym_customized_thing',
+        postInit: function (wym) {
+            deepEqual(wym._options.hrefSelector, '.wym_customized_thing');
+            deepEqual(WYMeditor.DEFAULT_DIALOG_OPTIONS.hrefSelector, startingDefault);
```
mightyiam: `strictEqual` instead of `deepEqual` in these two. Right?

src/test/unit/test.js
mightyiam: Some code comments to explain the test, perhaps?

src/wymeditor/core.js
```diff
@@ -541,251 +831,39 @@ jQuery.extend(WYMeditor, {
...
+            "Error locating skin '" + optionsWithDefaults.skin + "'."
+        );
+    }
+    optionsWithSkinOverrides = jQuery.extend(
+        {},
+        providedOptions,
+        WYMeditor._getSkinUIOverrides(providedOptions, skin)
+    );
 
-        boxHtml: String() +
```
mightyiam: I'm happy that all of this is out of here.

src/wymeditor/core.js
mightyiam: Let's make it a separate file.

src/wymeditor/core.js
mightyiam: Another one, below.

mightyiam:
Excuse me for being thorough and trying to understand this PR. Here go:

* [ ] [IE10 and older](https://github.com/wymeditor/wymeditor/pull/710#issuecomment-117104747)
* [ ] [A changelog entry change suggestion](https://github.com/wymeditor/wymeditor/pull/710/files#r33558443)
* [ ] Changelog entry ([some questions](https://github.com/wymeditor/wymeditor/pull/710/files#r29130697))
* [ ] [what is the default skin's name now?](https://github.com/wymeditor/wymeditor/pull/710/files#r33569653)
* [ ] [Use `strictEqual`](https://github.com/wymeditor/wymeditor/pull/710/files#r29133184)
* [ ] [Use `strictEqual`](https://github.com/wymeditor/wymeditor/pull/710/files#r33570831)
* [ ] [Code comments to explain test](https://github.com/wymeditor/wymeditor/pull/710/files#r33571038)
* [ ] [Improve code comment of `UI_COMPONENTS` and `UI_COMPONENT_SELECTORS` constants](https://github.com/wymeditor/wymeditor/pull/710/files#r29133679)
* [ ] [I don't understand what you mean](https://github.com/wymeditor/wymeditor/pull/710/files#r29134832)
* [ ] [Make default options a separate file](https://github.com/wymeditor/wymeditor/pull/710/files#r33571701)
* [ ] [More elegant options merging code](https://github.com/wymeditor/wymeditor/pull/710/files#r29136702)
* [ ] [`_getSkinUIOverrides` should be a non–static method that performs the merging as well?](https://github.com/wymeditor/wymeditor/pull/710/files#r29137540)
* [ ] [Use `Array.prototype.forEach`](https://github.com/wymeditor/wymeditor/pull/710/files#r33556376)

src/wymeditor/core.js
winhamwr: A separate file for a single constant?

src/wymeditor/core.js
winhamwr: > For starters, what are "the rest of the WYMeditor attributes?

[these](https://github.com/wymeditor/wymeditor/blob/master/src/wymeditor/core.js#L123).

src/test/unit/test.js
winhamwr: :+1:

src/test/unit/test.js
winhamwr: I read through it again and I'm not sure where comments would help. The name of the test succinctly explains what this is looking for and then the rest of the code has good variable names to explain what's being done/tested. Suggestions?

CHANGELOG.md
winhamwr: :+1:

src/test/unit/specific_feature_tests/skins.js
winhamwr: :+1:

src/wymeditor/core.js
winhamwr: I wrote it like this on purpose because it makes it very explicit what each `extend` is adding and how the precedence works. I think it provides more clarity than a single merge.

src/wymeditor/core.js
winhamwr: > this static method may be easier to understand if it also did the merging and not only provide the object for the merge.

I think limiting it to just providing the objects (doing exactly one thing) makes it easier to test and understand. It all supports the composability desired to perform the multi-step merge. Why have a function that does two things when it's useful doing one thing?

src/wymeditor/core.js
winhamwr: :+1:

src/wymeditor/core.js
winhamwr: :+1:

src/test/unit/specific_feature_tests/skins.js
winhamwr: default is the default, but the seamless skin doesn't look as nice for unit tests, since it causes the test results at the bottom to move around as content is added. I could be convinced that it's better to use the default skin and deal with the moving around, though.

src/test/unit/test.js
winhamwr: :+1:

winhamwr:
> Bang. Is this expected to work with all IE versions that we support?
> Testing now, the only IE version in which the height of the Iframe is growing, is IE11.

Well, shoot. IE8+ should work and they're working within PolicyStat on wymeditor [1.0.0-rc.1.patch3](https://github.com/PolicyStat/PolicyStat/blob/master/site_media/lib/wymeditor/wymeditor/jquery.wymeditor.js#L394). I either didn't add some necessary library or something in master changed since that version of WYMeditor.

> I suggest we release v1.0.0 without this.

I'd like to get this in as the default skin for v1.0.0. We're going to need to figure out why the skin isn't working before pushing to PolicyStat, anyway, and the current default skin is old and busted. This skin is a much better default for people going forward.

mightyiam:
I'll have to carry on with this issue when I'm home (14th). Incidentally, this is top priority task now, but it isn't feasible to test in all browsers on this laptop I got here.

winhamwr:
> I'll have to carry on with this issue when I'm home (14th).

Sounds like a plan. Hopefully your beefy desktop will make easy work of things upon your return. Have a safe trip home!

mightyiam:
In this branch, the current default skin, `'basic'`, is deleted.

The seamless skin is not suitable for the test suite because it takes impossibly longer to run.

mightyiam:
I'm going to try a new branch, where I

1. replace the logic within the seamless skin with [this module](https://github.com/davidjbradshaw/iframe-resizer)
1. try the speed in the test suites
 1. if it takes too long with this module, I add an option to disable resizing

PR #711, opened 2015-04-25
nativeEditRegistration._onAnyNativeEdit: fix object-history usage

mightyiam:
Update object-history

### DIFF
--- a/src/wymeditor/editor/native-edit-registration.js
+++ b/src/wymeditor/editor/native-edit-registration.js
@@ -51,7 +51,9 @@ WYMeditor.NativeEditRegistration.prototype._onAnyNativeEdit = function () {
         undoRedo = nativeEditRegistration.wym.undoRedo;
 
     // remove redo points
-    undoRedo.history.changesetsFore = [];
+    if (undoRedo.history) {
+        undoRedo.history.forgetAllForward();
+    }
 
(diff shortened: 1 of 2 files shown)


src/wymeditor/editor/native-edit-registration.js
```diff
@@ -51,7 +51,9 @@ WYMeditor.NativeEditRegistration.prototype._onAnyNativeEdit = function () {
         undoRedo = nativeEditRegistration.wym.undoRedo;
 
     // remove redo points
-    undoRedo.history.changesetsFore = [];
```
mightyiam: This private property was gone in object-history v3.0.0.

This would have been caught had we had proper test coverage for undo.

src/wymeditor/editor/native-edit-registration.js
```diff
@@ -51,7 +51,9 @@ WYMeditor.NativeEditRegistration.prototype._onAnyNativeEdit = function () {
...
+    if (undoRedo.history) {
+        undoRedo.history.forgetAllForward();
```
mightyiam: The [new API–foo](https://github.com/PolicyStat/object-history#forgetallforward).

mightyiam:
Thank you!

PR #724, opened 2015-07-22
Simplify the seamless skin's example

### LINKED ISSUE #723: Seamless skin example broken

mightyiam:
Fixes #723

### DIFF
--- a/src/examples/20-seamless.html
+++ b/src/examples/20-seamless.html
@@ -43,7 +43,12 @@
     <script type="text/javascript">
         jQuery(function() {
             jQuery('.wymeditor').wymeditor({
-              html: '<p>Some initial content... Type in a few paragraphs and see the editor grow in height.</p>',
+              html: [
+                '<p>Some initial content...</p>',
+                '<p>And an image for asynchronously added size:</p>',
+                '<img src="https://upload.wikimedia.org/wikipedia/commons/a/ac/Flag_of_Indiana.svg" alt="Flag of Indiana">',
+                '<p>Type in a few paragraphs and see the editor grow in height.</p>'
+              ].join(''),
                 skin: 'seamless',
                 // We already included the seamless skin's javascript so that we could
                 // use this constant
--- a/src/examples/20-seamless.html
+++ b/src/examples/20-seamless.html
@@ -18,10 +18,10 @@
             The seamless skin uses the height of your textarea as the minimum
             editor height.
         -->
-        <textarea class="wymeditor" style="min-height: 600px"></textarea>
+        <textarea class="wymeditor" style="min-height: 300px"></textarea>
         <input type="submit" class="wymupdate" />
     </form>
-    <div class="more-content">The toolbar will stop floating here.<div>
+    <p style="height: 5000px">The floating toolbar will stop here.</p>
 
     <!-- build:js vendor/jquery/jquery.js -->
     <!-- endbuild -->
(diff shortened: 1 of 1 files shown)


src/examples/20-seamless.html
```diff
@@ -42,34 +42,13 @@
...
-                    , 'src="../../skins/wymeditor_icon.png'
-                    , '?cache_break=' + i + '"'
-                    , ' />'
-                ].join('');
-            }
-            html += '<p>After the images</p>';
-
             jQuery('.wymeditor').wymeditor({
-                html: html,
+              html: '<p>Some initial content... Type in a few paragraphs and see the editor grow in height.</p>',
```
mightyiam: This was very confusing, because the initial document content was  identical to the "more-content" that was filled below the editor. Now it does not confuse.

src/examples/20-seamless.html
```diff
@@ -18,10 +18,10 @@
             The seamless skin uses the height of your textarea as the minimum
             editor height.
         -->
-        <textarea class="wymeditor" style="min-height: 600px"></textarea>
+        <textarea class="wymeditor" style="min-height: 300px"></textarea>
```
mightyiam: 600 was too high. It required too many new paragraphs until it starts increasing height.

src/examples/20-seamless.html
```diff
@@ -18,10 +18,10 @@
...
         <input type="submit" class="wymupdate" />
     </form>
-    <div class="more-content">The toolbar will stop floating here.<div>
+    <p style="height: 5000px">The floating toolbar will stop here.</p>
```
mightyiam: Clarification.

mightyiam: And also, instead of all of the below loop code and then a two (why two?) `jQuery.append`s, just give this some height with CSS.

src/examples/20-seamless.html
```diff
@@ -42,34 +42,13 @@
...
     <script type="text/javascript">
         jQuery(function() {
-            var stringToRepeat = '<p>Hello, World!<\/p>',
-                html = '';
-
-            for (var i = 0; i < 2; i++) {
-                html += stringToRepeat;
-            }
-            for (var i = 0; i < 5; i++) {
-                // A working image, relative to the iframe, with no height
```
mightyiam: I'm not sure why an image. Was it to test something in particular?

mightyiam: Perhaps the image is for an asynchronous increase in height.

src/examples/20-seamless.html
```diff
@@ -42,34 +42,18 @@
...
-                ].join('');
-            }
-            html += '<p>After the images</p>';
-
             jQuery('.wymeditor').wymeditor({
-                html: html,
+              html: [
+                '<p>Some initial content...</p>',
+                '<p>And an image for asynchronously added size:</p>',
+                '<img src="https://upload.wikimedia.org/wikipedia/commons/a/ac/Flag_of_Indiana.svg" alt="Flag of Indiana">',
```
mightyiam: Added an image, as well.

src/examples/20-seamless.html
jlward: Does the editor allow giving a `p` tag height? If not I would suggest we don't do that. Test/examples should be possible in the editor.

src/examples/20-seamless.html
jlward: I recall @winhamwr having a few problems with the seamless editor and images.

src/examples/20-seamless.html
mightyiam: This does not go in the editor's document.

This is for something else.

The seamless skin has a "floating toolbar". It means that as you scroll down past it, it sticks to the top of the window, or something. And It doesn't get past the document view.

To be able to scroll down, there usually needs to be some content, below the editor. And this is what this `p` is for. That is why it has height.

src/examples/20-seamless.html
winhamwr: > Perhaps the image is for an asynchronous increase in height.

Exactly. It's loaded after the original height is calculated.

src/examples/20-seamless.html
winhamwr: :+1:

mightyiam:
Thank you!

PR #730, opened 2015-08-13
`insert` inconsistent across browsers

mightyiam:


src/wymeditor/editor/base.js
```diff
@@ -2032,18 +2032,15 @@ WYMeditor.editor.prototype.paste = function (str) {
 WYMeditor.editor.prototype.insert = function (html) {
     // Do we have a selection?
     var wym = this,
-        selection = wym.selection(),
+        selection = wym.hasSelection() ? wym.selection() : false,
```
mightyiam: Get selection only if it exists.

src/wymeditor/editor/base.js
```diff
@@ -2032,18 +2032,15 @@ WYMeditor.editor.prototype.paste = function (str) {
...
         range,
         node;
-    if (selection.focusNode !== null) {
+    if (selection) {
```
mightyiam: Proper selection existence check now.

src/wymeditor/editor/base.js
```diff
@@ -2032,18 +2032,15 @@ WYMeditor.editor.prototype.paste = function (str) {
...
         // Overwrite selection with provided html
         range = selection.getRangeAt(0);
         node = range.createContextualFragment(html);
         range.deleteContents();
         range.insertNode(node);
-    } else {
-        // Fall back to the internal paste function if there's no selection
-        wym.paste(html);
```
mightyiam: This caused the IE bug. IE, when there was no selection, would `paste`.

`paste`, in turn, when there's no selection, inserts at the top of the document.

Not sure why this fallback was [ever made](https://github.com/wymeditor/wymeditor/commit/f86ac09a486b9c0c330e6ef16ed493207c5f2825).

I'm willing to part with it.

src/wymeditor/editor/trident-pre-7.js
```diff
@@ -115,25 +115,6 @@ WYMeditor.WymClassTridentPre7.prototype._saveCaret = function () {
...
-        try {
-            // Overwrite selection with provided html
-            range.pasteHTML(html);
-        } catch (e) {}
-    } else {
-        // Fall back to the internal paste function if there's no selection
-        wym.paste(html);
-    }
-};
-
```
mightyiam: The general `insert` works just as well in IE8 through 10 (11 is not this file).

src/wymeditor/editor/base.js
kylegibson: Is the ternary operator consistent with wymeditor javascript coding style guidelines? I know we try to avoid it in PolicyStat code, since it doesn't improve code readability, and frequently reduces it. I don't have a strong preference on this.

kylegibson:
Looks great. Beautiful JavaScript.

src/wymeditor/editor/base.js
mightyiam: [Style guide](http://wymeditor.readthedocs.org/en/latest/wymeditor_development/coding_standard.html)
doesn't mention it and it refers to the Opera style guide, right at the end, which does suggest use of it. And it is very common in modern JS.

mightyiam:
Thank you, @kylegibson.

PR #731, opened 2015-08-14
Popups in IE need to load a document

mightyiam:
If `window.open` in IE does not load a document,
the parent window may not be allowed to control the child,
due to some "same-origin domain policy" security issue.

This results in entirely unusable popup windows.

This may be the cause of the following from the customer support room:
screenshot_2015-08-15_01-53-35

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -360,6 +360,13 @@ module.exports = function (grunt) {
                             "iframe/{,*/}*.{html,js,css,png,jpg,jpeg,gif,eot,ttf,woff}"
                         ]
                     },
+                    // Popup document
+                    // https://github.com/wymeditor/wymeditor/issues/731
+                    {
+                        cwd: '<%= yeoman.app %>/wymeditor',
+                        dest: '<%= yeoman.dist %>/wymeditor',
+                        src: "popup.html"
+                    },
                     // Non-Javascript skin components
                     // The javascript is included in jquery.wymeditor.js
                     {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -363,9 +363,8 @@ module.exports = function (grunt) {
                     // Popup document
                     // https://github.com/wymeditor/wymeditor/issues/731
                     {
-                        cwd: '<%= yeoman.app %>/wymeditor',
-                        dest: '<%= yeoman.dist %>/wymeditor',
-                        src: "popup.html"
+                        dest: '<%= yeoman.dist %>/wymeditor/popup.html',
+                        src: '<%= yeoman.app %>/wymeditor/popup.html'
                     },
                     // Non-Javascript skin components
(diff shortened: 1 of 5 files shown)


src/wymeditor/popup.html
```diff
@@ -0,0 +1 @@
+<!-- Workaround for IE https://github.com/wymeditor/wymeditor/issues/731 -->
```
mightyiam: So we make a blank document.

src/wymeditor/editor/dialogs.js
```diff
@@ -86,11 +107,7 @@ WYMeditor.editor.prototype.dialog = function (
         ].join(",");
     }
 
-    wDialog = window.open(
-        '',
-        "wymDialogWindow",
-        dialogWindowFeatures
-    );
+    wDialog = wym._openPopupWindow(dialogWindowFeatures)
```
mightyiam: We make this into a method.

src/wymeditor/editor/dialogs.js
```diff
@@ -1,5 +1,26 @@
...
+    Opens a popup window, provided the `windowFeatures`.
+*/
+WYMeditor.editor.prototype._openPopupWindow = function (windowFeatures) {
+    var wym = this,
+        popup,
+        basePath = wym._options.basePath;
+
+    popup = window.open(
+        // https://github.com/wymeditor/wymeditor/issues/731
+        jQuery.browser.msie ? basePath + 'popup.html' : '',
```
mightyiam: All other browsers do no have this issue thus are fine with a blank string here.

src/wymeditor/core.js
```diff
@@ -936,7 +936,7 @@ WYMeditor._computeWymPath = function () {
     the wymeditor base file. This path is used as the basis for loading:
     * Language files
     * Skins
-    *
+    * The popup document https://github.com/wymeditor/wymeditor/issues/731
```
mightyiam: There was an asterisk here, waiting for me.

Gruntfile.js
```diff
@@ -360,6 +360,13 @@ module.exports = function (grunt) {
                             "iframe/{,*/}*.{html,js,css,png,jpg,jpeg,gif,eot,ttf,woff}"
                         ]
                     },
+                    // Popup document
+                    // https://github.com/wymeditor/wymeditor/issues/731
+                    {
+                        cwd: '<%= yeoman.app %>/wymeditor',
+                        dest: '<%= yeoman.dist %>/wymeditor',
+                        src: "popup.html"
+                    },
```
mightyiam: This includes the popup document in the distribution.

mightyiam: I tested that the distribution works, as well.

jlward:
Looks good.

mightyiam:
Thank you, @jlward.

PR #733, opened 2015-08-15
Seamless skin: slight space saving in toolbar

mightyiam:


jlward:
Does this need an update note? (Change log entry). Can I see a screen shot of the before and after state?

jlward:
Looks good.

mightyiam:
@jlward I do not think a changelog entry is required.

Before
screenshot_2015-08-24_11-48-15

After
screenshot_2015-08-24_11-48-50

PR #735, opened 2015-08-22
Hard code seamless skin height strategy in IE<11

mightyiam:
In the examples there is no issue.

In a specific integration, in IE<11, the height strategy for Chrome/FF is chosen, instead of the IE<11 strategy.

It seems more reliable to hard-code the strategy to the browser version, than the logic we currently have in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ WYMeditor.
 
 *release-date* TBD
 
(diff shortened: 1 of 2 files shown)


winhamwr:
:+1:

winhamwr:
Sucks that the detection doesn't work, but new version of IE less than 11 aren't going to come out, so this seems pretty safe.

mightyiam:
Thank you!

PR #739, opened 2015-09-09
Selection across table cells broken in Firefox

mightyiam:
[Here](https://github.com/wymeditor/wymeditor/blob/1b072bb3f86c925cbf9ca4dfd8a2004613fc374f/src/wymeditor/editor/base.js#L3021) in firefox, we get only a partial array of nodes. We do not get the entire expected selection, as is shown in GUI.

We get only the selected nodes from the logically first, selected table cell.

### DIFF
--- a/src/wymeditor/editor/base.js
+++ b/src/wymeditor/editor/base.js
@@ -3018,7 +3018,10 @@ WYMeditor.editor.prototype._getSelectedNodes = function () {
         return false;
     }
 
-    selectedNodes = selection.getRangeAt(0).getNodes();
+    selectedNodes = selection.getAllRanges()
+        .reduce(function(nodes, range) {
+            return nodes.concat(range.getNodes());
+        }, [])
     return selectedNodes;
 };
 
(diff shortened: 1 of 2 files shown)


src/wymeditor/editor/base.js
```diff
@@ -3018,7 +3018,10 @@ WYMeditor.editor.prototype._getSelectedNodes = function () {
         return false;
     }
 
-    selectedNodes = selection.getRangeAt(0).getNodes();
+    selectedNodes = selection.getAllRanges()
+        .reduce(function(nodes, range) {
+            return nodes.concat(range.getNodes());
+        }, [])
```
mightyiam: Firefox now returns a selection with three ranges, one for each cell.

This new code works in all browsers.

It is funny code, for me. I had no trouble writing it and it was fun. But, I don't like coming across code like this. It seems hard to read. I know that for others it may seem very easy to read. Is this a matter of experience?

kylegibson:
Is there no way for this behavior to be tested?

mightyiam:
With WebDriver it should be possible, but I'm not actively using WebDriver in the WYMeditor process.

With JavaScript and the selection API, either through Rangy or raw, I'm not sure.
I tried and got stuck somewhere. Not worth it.

kylegibson:
Otherwise this looks good

mightyiam:
Thank you!

PR #740, opened 2015-09-16
Fix package.json

mightyiam:
* [Syntax](https://docs.npmjs.com/files/package.json#license)
* [SPDX Licenses](https://spdx.org/licenses/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,16 +24,7 @@
     "name": "Jean-Francois Hovinne",
     "email": "jf.hovinne@wymeditor.org"
   },
-  "licenses": [
-    {
-      "type": "MIT",
-      "url": "http://mit-license.org/"
-    },
-    {
-      "type": "GPLv2",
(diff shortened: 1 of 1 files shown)


caffodian:
Licenses change looks good.  However I think the repository field needs to be fixed too.  

I tried passing it into http://package-json-validator.com/  as well as this other NPM validator that's on npm, they both say that the repository field should be an object or a string (not a single item array with an object inside.)  This checks out with what the npm docs say [here](https://docs.npmjs.com/files/package.json#repository) 

Here is the output of the validator i tried, if you are curious:

```
{
  "valid": false,
  "errors": [
    "Type for field repository, was expected to be string or object, not object"
  ],
  "warnings": [
    "Missing recommended field: contributors"
  ],
  "recommendations": [
    "Missing optional field: dependencies"
  ]
}
````

mightyiam:
Thank you, @caffodian

PR #741, opened 2015-09-19
Allow wrap/unwrap toggle across elements

mightyiam:
Initial report suggests an issue. Not sure whether or how much this was possible in previous versions.

### DIFF
--- a/src/test/unit/specific_feature_tests/exec_toggles.js
+++ b/src/test/unit/specific_feature_tests/exec_toggles.js
@@ -1,4 +1,4 @@
-/* jshint evil: true */
+/* jshint evil: true, maxlen: 100 */
 /* global
     manipulationTestHelper,
     test,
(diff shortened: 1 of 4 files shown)


src/test/unit/specific_feature_tests/exec_toggles.js
```diff
@@ -318,11 +318,11 @@ test("IE unwraps partially subscript selection", function () {
     );
 });
 
-function doesntWrapAcrossRootContainers(command) {
-    var noChangeHtml = "<p>Foo</p><p>Bar</p>";
+function wrapsAcrossRootContainers(command) {
+    var tagName = getBrowserTagname(command);
     manipulationTestHelper({
-        startHtml: noChangeHtml,
-        prepareFunf: function (wymeditor) {
```
mightyiam: Had a typo here previously! Bad test! How do I avoid this in JavaScript?

kylegibson: `manipulationTestHelper` could do a validation on the object it receives, and if there are any keys missing, or unexpected keys passed it, it aborts.

kylegibson:
The code looks good. However, I notice that the implementation and tests focus on "root-level" containers, when the bug report just mentions items, presumably non-root level containers.

1. Does this fix work with non-root level containers?
2. Are there any tests for this on non-root level containers?
3. The previous implementation did not allow wrap/unwrap on root level containers. What was the rationale for that previous behavior, and why are we changing it now? (I could see case where non-root level containers would be wrap/unwrap, but not root level)

winhamwr:
Tracing this back to #632 makes me feel more comfortable with this fix.

winhamwr:
I think the answer to Kyle's questions are:

1. I think yes, because the [selectedContainer()](https://github.com/wymeditor/wymeditor/blob/master/src/wymeditor/editor/base.js#L849) check was disallowing the wrap acorss any sibling-level selection (or anything other than a selection from a parent to a child). 
2. No. I think at least one test across e.g. sibling items in a sub-list would be useful.
3. I looked through #632 and I can't figure out for the life of me why we thought it would be ideal to not allow people to bold across root level containers. Do you remember @mightyiam?

mightyiam:
1. What Wes wrote.
2. I'll make such a test.
3. It is not easy to understand the implications of changes in such a complex UI (most of which is whatever the current browser decides to do).
  And I'm not really using the editor for anything so, I think I need to eat my own :dog2: food.

src/test/unit/specific_feature_tests/exec_toggles.js
```diff
@@ -338,67 +338,187 @@ function doesntWrapAcrossRootContainers(command) {
...
-test("Doesn't unwrap superscript across root containers", function () {
-    doesntUnwrapAcrossRootContainers(WYMeditor.EXEC_COMMANDS.SUPERSCRIPT);
+test("unwraps superscript from inside `li` to child of sibling `li`", function () {
+    unwrapsFromInsideLiToChildOfSiblingLi(WYMeditor.EXEC_COMMANDS.SUPERSCRIPT);
 });
 
-test("Doesn't unwrap subscript across root containers", function () {
-    doesntUnwrapAcrossRootContainers(WYMeditor.EXEC_COMMANDS.SUPERSCRIPT);
+test("unwraps subscript from inside `li` to child of sibling `li`", function () {
+    unwrapsFromInsideLiToChildOfSiblingLi(WYMeditor.EXEC_COMMANDS.SUBSCRIPT);
```
mightyiam: This is a mess. Perhaps in the commit, it is easier to see that I added some more tests.

mightyiam:
New tests passed on supported browsers.

winhamwr:
Other than the jshint problems, looks good!

mightyiam:
Now I know why you came down to that multiline style, Wes.

VINTproYKT:
Nicely done! I think it will be more convenient to apply some styles directly on paragraphs and other  blocks. This is exactly what will reveal (or improve) a concept of WYSIWYM. After I done with my basic model of app, I'll try to show what I mean.

PR #744, opened 2015-09-22
Fix dist's license string

mightyiam:
This was broken in #740. This is the difference it makes (top of dist file):
screenshot_2015-09-22_21-38-24

Without this change, it shows
```
Licenced 
```
(no license is shown).


winhamwr:
:+1: 

Looks good.

PR #751, opened 2015-10-08
WYMeditor.editor.hasSelection non-IE browsers fix + correcting Russian

VINTproYKT:
Testing my first pull request.


winhamwr:
Thanks for the pull request!

I'm not sure I understand what's being fixed here (other than I assume correcting some Russian). What happens in IE8 right now, and what happens after your change?

-Wes

VINTproYKT:
@winhamwr Oh, I just forgot to explain.

I have error in Chromium (maybe check it in other browsers), addressing to `rangy.isSelectionValid`, that throws "`undefined` is not a function". And @mightyiam just meant that this is only for IE v8 and before, so I isolated this piece of code from other browsers.

mightyiam:
@VINTproYKT I appreciate the contribution.

I do remember that this was an issue only in IE8.

But I don't see the same issue that you do. Using Chrome I do have `rangy.isSelectionValid`. In Chrome, it is defined [here](https://github.com/wymeditor/wymeditor/blob/master/src/wymeditor/rangy/rangy-core.js#L2470) and there is another definition of it for perhaps other browsers just above that.

So, I'd like to know why you don't have it defined. Can you please debug and see what causes it? Perhaps it gets [here](https://github.com/wymeditor/wymeditor/blob/master/src/wymeditor/rangy/rangy-core.js#L2474)?

Oh, and from now on please provide separate pull requests for non-related issues (translation, here).

PR #747, opened 2015-10-08
Image resizing

mightyiam:
## Spec

* [x] Allows resizing images easily and intuitively
* [x] Maintains original image proportions

## After Merging
* [ ] File issue: in IE 8 through 11 SVGs are not scaled but cropped
* [ ] File issue: Drag and drop of images disabled in IE9.


mightyiam:
Wonderful and smooth in Chrome and Firefox, so far. A few issues remain. I've listed them in the description.

The first item, the image cropping instead of scaling in IE is the only thing I don't have a clue on.

mightyiam:
Perhaps cause a reload of the image `onmouseup`. Yeah, that should do it, I guess.

src/examples/20-seamless.html
```diff
@@ -51,7 +51,7 @@
               html: [
                 '<p>Some initial content...</p>',
                 '<p>And an image for asynchronously added size:</p>',
-                '<img src="https://upload.wikimedia.org/wikipedia/commons/a/ac/Flag_of_Indiana.svg" alt="Flag of Indiana">',
+                '<p><img src="https://upload.wikimedia.org/wikipedia/commons/thumb/a/ac/Flag_of_Indiana.svg/500px-Flag_of_Indiana.svg.png" alt="Flag of Indiana"></p>',
```
mightyiam: Do not use an SVG. They can't be resized in IE. Instead, they are cropped.

.jshintrc
```diff
@@ -51,7 +51,7 @@
     "jquery"        : true,
 
     "nomen"         : false,
-    "onevar"        : true,
+    "onevar"        : false,
```
mightyiam: Blah. I don't like this style and [it was removed in JSHint. Notice you can't find it](http://jshint.com/docs/options/).

mightyiam: I :heart_decoration: [js-standard-style](https://github.com/feross/standard)

Gruntfile.js
```diff
@@ -274,6 +274,7 @@ module.exports = function (grunt) {
                     "<%= yeoman.app %>/wymeditor/editor/base.js",
                     "<%= yeoman.app %>/wymeditor/editor/" +
                         "document-structure-manager.js",
+                    "<%= yeoman.app %>/wymeditor/editor/image-handler.js",
```
mightyiam: Concatenate me this :kiss:

src/test/load-src.dev.js
```diff
@@ -21,6 +21,7 @@ function loadWymSrc(srcPath, extraRequirements, jqueryVersion) {
             srcPath + 'wymeditor/editor/dialogs.js',
             srcPath + 'wymeditor/editor/base.js',
             srcPath + 'wymeditor/editor/document-structure-manager.js',
+            srcPath + 'wymeditor/editor/image-handler.js',
```
mightyiam: Load me this in dev :kiss:

src/test/unit/index.html
```diff
@@ -70,6 +70,7 @@
             "specific_feature_tests/undo_redo.js",
             "specific_feature_tests/selection.js",
             "specific_feature_tests/images.js",
+            "specific_feature_tests/image-resize-handle.js",
```
mightyiam: Test me this :kiss:

src/test/unit/specific_feature_tests/images.js
```diff
@@ -188,99 +185,32 @@ test("Returns an image when it is exclusively selected", function () {
 
 module("images-selection", {setup: prepareUnitTestModule});
 
-test("Image is selected via `mouseup` in non pre-7 Trident", function () {
+test("Image is selected via `click`", function () {
```
mightyiam: This one is much simpler now.

src/test/unit/specific_feature_tests/xml_parser.js
```diff
@@ -934,6 +936,18 @@ test("Class removal is case insensitive", function () {
     WYMeditor.CLASSES_REMOVED_BY_PARSER = defaultClassesRemovedByParser;
 });
 
+module("XmlParser-remove_style_attribute", {setup: prepareUnitTestModule});
+
+test("Style attribute is removed from images", function () {
```
mightyiam: Because we use style on the image for UI.

src/wymeditor/core.js
```diff
@@ -1087,7 +1090,7 @@ jQuery.fn.parentsOrSelf = function (selector) {
     }
 };
 
-// String & array helpers
+// Various helpers
```
mightyiam: Because I added a helper that isn't about strings.

src/wymeditor/core.js
```diff
@@ -1141,6 +1144,20 @@ WYMeditor.Helper = {
             }
         }
         return null;
+    },
+
+    // naively returns all event types
+    // of the provided an element
+    // according the its property keys that
+    // begin with 'on'
+    getAllEventTypes: function (elem) {
```
mightyiam: Helpful for debugging!!!

src/wymeditor/editor/base.js
```diff
@@ -398,6 +398,8 @@ WYMeditor.editor.prototype._afterDesignModeOn = function () {
 
     wym.nativeEditRegistration = new WYMeditor.NativeEditRegistration(wym);
 
+    wym.ih = new WYMeditor.ImageHandler(wym);
```
mightyiam: New awesomeness in town.

src/wymeditor/editor/base.js
```diff
@@ -3812,10 +3814,6 @@ WYMeditor.editor.prototype._afterInsertTable = function () {
 WYMeditor.editor.prototype._listen = function () {
     var wym = this;
 
-    wym.$body().bind("mouseup", function (e) {
-        wym._mouseup(e);
-    });
```
mightyiam: Was used for selecting images. No more.

mightyiam: Is done using clicks on the ImageHandler now.

src/wymeditor/editor/base.js
```diff
@@ -3857,14 +3855,8 @@ WYMeditor.editor.prototype._selectSingleNode = function (node) {
...
-    selection.setSingleRange(nodeRange);
-};
 
-WYMeditor.editor.prototype._mouseup = function (evt) {
-    var wym = this;
-    if (evt.target.tagName.toLowerCase() === WYMeditor.IMG) {
-        wym._selectSingleNode(evt.target);
-    }
+    selection.setSingleRange(nodeRange);
 };
```
mightyiam: Confusing diff here. Watch out. The poet's intention is removal of the `_mouseup` handler.

That is because we do not use it to select images any more. We use the `ImageHandler`.

src/wymeditor/editor/gecko.js
```diff
@@ -19,16 +19,14 @@ WYMeditor.WymClassGecko.NEEDS_CELL_FIX = parseInt(
 
 WYMeditor.WymClassGecko.prototype._docEventQuirks = function () {
     var wym = this;
+    var $doc = jQuery(wym._doc);
```
mightyiam: Cache it.

src/wymeditor/editor/gecko.js
```diff
@@ -19,16 +19,14 @@ WYMeditor.WymClassGecko.NEEDS_CELL_FIX = parseInt(
...
-    jQuery(wym._doc).bind("click", function (evt) {
-        wym._click(evt);
-    });
-    jQuery(wym._doc).bind("focus", function () {
+    $doc.keyup(wym._keyup.bind(wym));
+    $doc.focus(function () {
+        // not providing _onBodyFocus here because it doesn't exist yet
         wym.undoRedo._onBodyFocus();
     });
+    $doc.click(wym._click.bind(wym));
```
mightyiam: These are just style changes.

mightyiam: Unrelated, you could say.

src/test/unit/specific_feature_tests/image-resize-handle.js
```diff
@@ -0,0 +1,130 @@
...
+                .mousemove()
+                .remove();
+            wymeditor.$body().find('.wym-resize-handle').mousemove();
+        },
+        additionalAssertionsFunc: function (wymeditor) {
+            expectOneMore();
+            strictEqual(wymeditor.$body().find('div').length, 0);
+        }
+    });
+});
```
mightyiam: There can be many more tests, but this is good enough for the benefit we're getting.

src/wymeditor/editor/trident-7.js
```diff
@@ -51,9 +51,6 @@ WYMeditor.WymClassTrident7.prototype._docEventQuirks = function () {
     jQuery(wym._doc).bind("keyup", function (evt) {
         wym._keyup(evt);
     });
-    jQuery(wym._doc).bind("click", function (evt) {
-        wym._click(evt);
-    });
```
mightyiam: This is unrelated and was completely useless.

mightyiam: This was as if we had a Trident 7 specific `click` handler. But we don't

src/wymeditor/parser/xhtml-validator.js
```diff
@@ -21,13 +21,21 @@ WYMeditor.XhtmlValidator = {
             "attributes":[
             "class",
             "id",
-            "style",
```
mightyiam: I've removed the `style` attribute from being valid on all elements except those listed under `except`, above.

src/wymeditor/parser/xhtml-validator.js
```diff
@@ -21,13 +21,21 @@ WYMeditor.XhtmlValidator = {
...
         },
+        "styleAttr":
+        {
+            "except":[
+            "img"
+            ],
+            "attributes":[
+            "style"
+            ]
+        },
```
mightyiam: And then I added the `style` attribute to be valid on all elements except `img`.

.jshintrc
caffodian: lol...of course there's already a fork with semicolons :p

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+ *
+ * Dragging and dropping of images is disabled.
+ * See the `_isImgDragDropAllowed` function.
+ *
+ * ## IE8-11 Shenanigans
+ *
+ * SVG images are not scaled.
+ * They are cropped. That's right.
+ * And applying style to them does not help,
+ * as well. Ideas are welcome.
```
caffodian: that sucks, wish I knew more about SVG to help.  I know there are weird quirks between img & object and the other ways of embedding svg, but I don't know the specifics

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+ *
+ * ## General Shenanigans (http://i.imgur.com/wbQ6U5C.jpg)
+ *
+ * This module is covered by only a few basic tests
+ * so any change must be meticulously manually tested
+ * in all the supported browsers
+ * by psychologically stable individuals.
```
caffodian: by nature, this eliminates anyone who actually works on ani n-browser editor :p

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+};
+
+WYMeditor.ImageHandler._isImgDragDropAllowed = function () {
+    var browser = jQuery.browser;
+    if (browser.msie) {
+        if (browser.versionNumber === 9) {
+            // dragging and dropping seems to not consistently work.
+            // the image would only some times get picked up by the mouse drag attempt.
+            // to prevent confusion
+            return false;
```
caffodian: yet it works in 8?  weird.

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+
+    ih._$resizeHandle
+        .css('font-weight', 'bold')
+        .html(indication);
+
+    // ideally, the above indication text would remain
+    // until the image is no longer selected.
+    // since it is not easy to detect when that happens,
+    // the indication text is replaced with the initial text
+    // after a short short.
```
caffodian: short wait?

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+        var originalHeight = $img.attr('height');
+        var originalWidth = $img.attr('width');
+        dimensionsRatio =  originalWidth / originalHeight;
+        $img.data('DimensionsRatio', dimensionsRatio);
+    }
+
+    // calculate the new dimensions
+    var startHeight = $img.data('StartHeight');
+    var newHeight = startHeight - ih._startMouseY + currentMouseY;
+    newHeight = newHeight > 0 ? newHeight : 0;
```
caffodian: is it intended to set it to 0 in this case?  or should it be 1?

caffodian: i guess you store the original ratio anyway.  so temporarily setting it to 0x0 might be okay.

there might be some extremely stupid cases with really really wide/tall images.  I'm not sure if we should save the user from themselves in that case.

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+    var ih = this;
+    var $img = ih._$currentImg;
+
+    var dimensionsRatio = $img.data('DimensionsRatio');
+
+    if (!dimensionsRatio) {
+        // in order to prevent dimensions ratio corruption
+        var originalHeight = $img.attr('height');
+        var originalWidth = $img.attr('width');
+        dimensionsRatio =  originalWidth / originalHeight;
```
caffodian: degenerate case where with original height=0.  not sure if worth handling.

src/wymeditor/editor/image-handler.js
```diff
@@ -0,0 +1,681 @@
...
+        // that it would be hard to mouse over it
+        // in order to make the resize handle appear.
+        // in that case (an arbitrary number of pixels)
+        // leave the padding on, as it will allow
+        // easy mouse over the image,
+        // even when the image is 0 in size
+        ih._$currentImg.attr('height') >= 16 &&
+        ih._$currentImg.attr('width') >= 16
+    ) {
+        ih._$currentImg.css({padding: 0, margin: 0});
```
caffodian: yay usability

caffodian:
I tried.  :)  some comments in resizehandler to consider, besides that it looks good.

src/wymeditor/editor/image-handler.js
mightyiam: :sweat_smile:

src/wymeditor/editor/image-handler.js
mightyiam: Yes. There's something going on under the hood which I didn't figure out.

I have just tried the same trick as with IE11 (selecting the image synchronously `onmousedown`. Didn't seem to help. I'm going to leave it like this.

src/wymeditor/editor/image-handler.js
mightyiam: Short moment. Thanks.

src/wymeditor/editor/image-handler.js
mightyiam: Yes, you understand correctly that 0x0 is alright if we store the original dimension (if not, they get lost real fast and end up as one–to–one).

This is meant to prevent positive dimensions turning into negative dimensions.

If the new height is smaller than 0―which happens often due to the mouse position―then set it to 0 (which is quite useless but harmless and feels like the natural UX).

I'm not sure why, though. It just seems a good idea.

I'm not sure why you think that the cases with really wide/tall images might be not addressed by this code. I think they are and that it doesn't matter.

I do see one issue and that is with very wide images, the text in the resize handle would be off the viewport.

I'm pretty sure it is not worth the extra effort to cover that case now.

src/wymeditor/editor/image-handler.js
mightyiam: Images with `height=0` have lost their original dimensions.

Those can be obtained by examining the source file's dimensions (assuming that it exists). This is done the same way (`$img.attr('height')`).

Another approach would be a button on the resize handle, which resets the image to its default dimensions by removing the `height` and `width` attributes.

mightyiam: Images with original zero height or width could never reach this code because they can't be `mousemove` over. So that's the end of this little edge case, I guess. Not worth the effort.

PR #754, opened 2015-10-28
Allow insert image and paste dialogs to open when selection is non-collapsed (not only when it is collapsed)

mightyiam:
You might call this a usability issue, but it too me 15 minutes to figure out why these dialogs worked sometimes and not others. I thought there was some deeper kind of bug. Has this always been the case?


caffodian:
looks good

mightyiam:
Thank you @caffodian

PR #755, opened 2015-10-28
Can't select text with mouse drag (in master)

mightyiam:
Most likely from #747's `mousedown` handler.

### DIFF
--- a/src/wymeditor/editor/image-handler.js
+++ b/src/wymeditor/editor/image-handler.js
@@ -424,10 +424,15 @@ WYMeditor.ImageHandler.prototype._startResize = function (startMouseY) {
 WYMeditor.ImageHandler.prototype._onMousemove = function (evt) {
     var ih = this;
 
+    // default action for this event may be a `mousedown` and `mousemove`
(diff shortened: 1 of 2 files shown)


src/wymeditor/editor/image-handler.js
```diff
@@ -424,10 +424,15 @@ WYMeditor.ImageHandler.prototype._startResize = function (startMouseY) {
 WYMeditor.ImageHandler.prototype._onMousemove = function (evt) {
     var ih = this;
 
+    // default action for this event may be a `mousedown` and `mousemove`
+    // with the intention to select text
+    // or to drag (for later dropping of) content
+    // so be careful about preventing default
```
mightyiam: This is the culprit.

Now it returns false only when it is surely not a text selection attempt.

caffodian: I should have looked closer at where these handlers were actually being bound.  good thing you played with the editor

caffodian:
:+1:

mightyiam:
Thank you!

PR #757, opened 2015-10-30
Trying to resize image results in zero size image

mightyiam:
Use `jQuery.fn.<height|width>()` for getting image size instead of `jQuery.fn.attr('<height|width>')`.

### DIFF
--- a/src/wymeditor/editor/image-handler.js
+++ b/src/wymeditor/editor/image-handler.js
@@ -417,7 +417,7 @@ WYMeditor.ImageHandler.prototype._startResize = function (startMouseY) {
     var ih = this;
 
     ih._startMouseY = startMouseY;
-    ih._$currentImg.data('StartHeight', ih._$currentImg.attr('height'));
+    ih._$currentImg.data('StartHeight', ih._$currentImg.height());
     ih._resizingNow = true;
 };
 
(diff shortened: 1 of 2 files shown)


src/wymeditor/editor/image-handler.js
```diff
@@ -519,8 +519,8 @@ WYMeditor.ImageHandler.prototype._resizeImage = function (currentMouseY) {
 
     if (!dimensionsRatio) {
         // in order to prevent dimensions ratio corruption
-        var originalHeight = $img.attr('height');
-        var originalWidth = $img.attr('width');
+        var originalHeight = $img.height();
+        var originalWidth = $img.width();
```
caffodian: ah, actual height vs. defined height on img?

caffodian:
:+1:

src/wymeditor/editor/image-handler.js
mightyiam: Yes. I should have been surprised that it actually worked.

Initially, not all images have `height` and `width` attributes. But it did work with those images, nonetheless.

Inside PolicyStat it doesn't, though. So perhaps it is a jQuery difference.

mightyiam:
Thank you

src/wymeditor/editor/image-handler.js
mightyiam: I should have explained this before you made your review.

PR #758, opened 2015-10-30
Changing `img` `src` attr causes wrong dimensions in image resize

mightyiam:
The image dialog should flag the image (whose `src` was updated) as such.

The `ImageHandler` should detect that and reset the `DimensionsRatio` data attribute.

### DIFF
--- a/src/wymeditor/editor/base.js
+++ b/src/wymeditor/editor/base.js
@@ -1803,6 +1803,30 @@ WYMeditor.editor.prototype.insertImage = function (attrs) {
 };
 
 /**
+    editor._updateImageAttrs
+    ========================
+
+    Updates provided `img`'s attributes with provided `attrs`.
+*/
+WYMeditor.editor.prototype._updateImageAttrs = function (img, attrs) {
+    var $img = jQuery(img);
+    if (attrs.src !== $img.attr('src')) {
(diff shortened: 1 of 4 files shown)


src/wymeditor/editor/base.js
```diff
@@ -1803,6 +1803,30 @@ WYMeditor.editor.prototype.insertImage = function (attrs) {
...
+        // height/width ratio is now most likely invalid
+        $img.data('DimensionsRatio', null);
+
+        // since the height/width ration is most likely different
+        // these proportions are most likely wrong.
+        // easiest solution is to remove them
+        // and let the user deal with
+        // the real size of the source image
+        // and he'll be able to resize it
+        $img.removeAttr('height').removeAttr('width');
```
caffodian: this might lead us in circles a bit, but I wonder if it might be worth keeping the height/width, if the new dimensions are the same?  this might lead to usability improvements in some cases (for example, if you are swapping out a logo or diagram and want it to take up the same amount of space)

mightyiam: That would be ideal.

WYMeditor is not a modern, forward facing editor, any more. So we won't be going the extra mile with a lot of things.

mightyiam:
Thank you!

PR #759, opened 2015-10-30
Click on selection does not collapse it into caret

mightyiam:
Back in `v1.0.7` there's no such issue.

Most likely a default action prevented in some handler.


caffodian:
:+1: yay regressions

mightyiam:
Thank you @caffodian

PR #760, opened 2015-10-30
Do not suggest image drag and drop in IE8

mightyiam:
Image dragging and dropping in IE8 seems to be, like in IE9, unstable.

I didn't figure out a way to disallow it. Giving up.

### DIFF
--- a/src/wymeditor/editor/image-handler.js
+++ b/src/wymeditor/editor/image-handler.js
@@ -12,6 +12,10 @@
  * When IE8 is not longer supported,
  * `rem` could be used for more accurate UI element dimensions
  *
+ * Dragging and dropping of images is not suggested in the UI.
+ * See the `_isImgDragDropAllowed` function
+ * and the `_onImgMousedown` function.
+ *
  * ## IE9 Shenanigans
  *
  * Dragging and dropping of images is disabled.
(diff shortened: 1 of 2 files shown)


src/wymeditor/editor/image-handler.js
```diff
@@ -575,6 +579,7 @@ WYMeditor.ImageHandler.prototype._onImgMousedown = function (evt) {
     }
 
     // returning false here prevents drag of image
+    // except for in IE8
```
mightyiam: So this doesn't really prevent drag and drop. I didn't try much figuring out a way to prevent drag and drop. I think that the important thing is that it is not suggested in the UI.

mightyiam:
It was actually working well for a while. I'm not sure what caused image drag and drop to be unstable in IE8. I don't care all that much about it, though, because IE8 is almost out of hit points.

mightyiam:
This will be a post–mergem review.

src/wymeditor/editor/image-handler.js
caffodian: not really an IE8 person at all and just throwing random ideas here.  does doing anything in dragstart handler help?

mightyiam: I remember that `dragstart` is not triggered by dragging in `designMode` in IE. At least not in 8. Perhaps since 11, if at all.

PR #761, opened 2015-11-08
Image resize handle below end of iframe

mightyiam:
This happens when the image is at the end of the content or almost there.

## Adding bottom padding to the body of the iframe

So, the simplest solution is:

1. add a fixed (~30px) padding to the body of the iframe
1. trigger a `postBlockMaybeCreated` event along with the resize operation

This doesn't work in IE<11, seemingly because of [a bug in the seamless skin](https://github.com/wymeditor/wymeditor/blob/64938676c0238681600b05bbee3cea8cff381338/src/wymeditor/skins/seamless/skin.js#L405-L410).

By "doesn't work" I mean that while resizing, the iframe constantly increases in height.

So, to solve that, I'd have to debug that.

## An iframe-resizing library

The above makes me think, "hey, instead of debugging that bug, I'll drop in an iframe-resizing library, like [this one](http://davidjbradshaw.github.io/iframe-resizer/)".

## `contentEditable`

But, wait, the only reason we have an iframe is because we use `designMode`. Perhaps a similar effort can be invested in migrating to `contentEditable`.

`contentEditable` most likely means that this current bug will be very easy to fix.

## `padding-bottom` the image

So, adding bottom padding to the image seems to do the trick.

Of course, it is a bit more complex than that so take a look at the code.

### DIFF
--- a/src/wymeditor/editor/image-handler.js
+++ b/src/wymeditor/editor/image-handler.js
@@ -342,12 +342,23 @@ WYMeditor.ImageHandler.prototype._indicateOnResizeHandleThatImageIsSelected = fu
 
 WYMeditor.ImageHandler.prototype._placeResizeHandleOnImg = function (img) {
     var ih = this;
+    var wym = ih._wym;
     var IMAGE_PADDING = '0.8em';
     var $img = jQuery(img);
 
     ih._$currentImg = $img;
 
     ih._getCurrentImageMarker().insertAfter($img);
+
+    // the resize handle, prepended to the body in this way,
+    // can be removed from the body using DOM manipulation
+    // such as setting the content with the `html` method.
+    // so we place it there in case that occurred.
+    // this could be done conditionally
+    // but there is practically no performance hit so keeping it simple
+    ih._$resizeHandle.prependTo(ih._wym.$body());
+
+    ih._$resizeHandle.show();
 
     // colored padding around the image and the handle
     // visually marks the image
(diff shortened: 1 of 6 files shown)


src/wymeditor/editor/image-handler.js
```diff
@@ -349,6 +350,16 @@ WYMeditor.ImageHandler.prototype._placeResizeHandleOnImg = function (img) {
 
     ih._getCurrentImageMarker().insertAfter($img);
 
+    // the resize handle, prepended to the body in this way,
+    // can be removed from the body using DOM manipulation
+    // such as setting the content with the `html` method.
+    // so we place it there in case that occurred.
+    // this could be done conditionally
+    // but there is practically no performance hit so keeping it simple
+    ih._$resizeHandle.prependTo(ih._wym.$body());
```
mightyiam: This is moved from further below.

src/wymeditor/editor/image-handler.js
```diff
@@ -349,6 +350,16 @@ WYMeditor.ImageHandler.prototype._placeResizeHandleOnImg = function (img) {
...
+
+    ih._$resizeHandle.show();
```
mightyiam: This is also moved from further below.

src/wymeditor/editor/image-handler.js
```diff
@@ -367,21 +389,13 @@ WYMeditor.ImageHandler.prototype._placeResizeHandleOnImg = function (img) {
...
-    ih._$resizeHandle.prependTo(ih._wym.$body());
-
     // it is important that the resize handle's offset
     // is updated after the above style modification
     // adds top padding to the image
     // because that alters the image's outside height
     ih._correctResizeHandleOffsetAndWidth();
 
-    ih._$resizeHandle.show();
+    jQuery(wym.element).trigger(WYMeditor.EVENTS.postBlockMaybeCreated, wym);
```
mightyiam: The triggering of this event throughout this module is kind of hackish, because the name of the event is obviously not correct.

Other than that, it seems to do what we desire, which is that the seamless skin will recalculate its height.

src/wymeditor/editor/image-handler.js
```diff
@@ -540,6 +560,8 @@ WYMeditor.ImageHandler.prototype._resizeImage = function (currentMouseY) {
     $img.attr('width', newWidth);
 
     ih._correctResizeHandleOffsetAndWidth();
+
+    jQuery(wym.element).trigger(WYMeditor.EVENTS.postBlockMaybeCreated, wym);
```
mightyiam: Doesn't seem to be a performance hit.

mightyiam:
I ended up solving this by adding a bottom-padding on the image.

michaelgzoller:
looks good to me

mightyiam:
Thank you!

Issue #762, opened 2015-11-19
Table editing is not cool


mightyiam:
## Reproducing instructions

1. Try to edit a table
1. Try not to flip a table

Firefox's native inline table editing is off.

The table editing plugin included in WYMeditor makes some bad manipulations when cells span more than one row/column.

### Fixing the table editing plugin

The most reasonable technical solution in WYMeditor would be to fix (probably mostly rewrite) the table editing plugin, including unit tests.

### Enabling the native table editing mechanism where it exists

Enabling native table resizing in Gecko [may prove quite easy](https://github.com/wymeditor/wymeditor/blob/7317c80fd9a977f65f6bb96e542a76a456d03c69/src/wymeditor/editor/gecko.js#L149).

:-1: This alone will not solve the bugs that are in the table editing plugin.

:-1: It may be that only Firefox has this native feature.

:-1: It also introduces an issue where width attributes are added to table elements (right?)

mightyiam:
@winhamwr do we like solution number 2? Is it worth it?

winhamwr:
#2 is the right fix for the short-term, I think. It at least gets us back to where we were from a usability standpoint.

PR #763, opened 2015-11-21
Enable Firefox's native table editing

mightyiam:


src/wymeditor/editor/gecko.js
```diff
@@ -146,7 +146,8 @@ WYMeditor.WymClassGecko.prototype._designModeQuirks = function () {
     try {
         wym._doc.execCommand("styleWithCSS", '', false);
         wym._doc.execCommand("enableObjectResizing", false, false);
-        wym._doc.execCommand("enableInlineTableEditing", false, false);
```
mightyiam: This enables the by-default Firefox's native table editing thingo, which looks like this:
screenshot_2015-11-21_20-06-32
It works quite alright, it seems, during some tinkering.

mightyiam:
More discussion in #762.

winhamwr:
:+1: 

Looks good! Does Chrome have an equivalent?

mightyiam:
@winhamwr Chrome doesn't seem to have such a feature. Doesn't surprise me at all.
http://stackoverflow.com/questions/25372336/enableobjectresizing-and-enableinlinetableediting-in-webkit

mightyiam:
Thank you!

src/wymeditor/editor/gecko.js
Geetha2111: It doesn't support other browsers like chrome or safari. Is there any other alternative for this.?

mightyiam: Brah, WYMeditor is such an old editor. Why don't you use some more modern alternative?

PR #764, opened 2015-11-21
TablePlugin row/column remove: refrain from error when no selectedContainer

mightyiam:


src/wymeditor/plugins/table/jquery.wymeditor.table.js
```diff
@@ -138,22 +138,38 @@ TableEditor.prototype.bindEvents = function () {
     jQuery(wym._box).find(
         tableEditor._options.sAddRowButtonSelector
     ).click(function () {
-        return tableEditor.addRow(wym.selectedContainer());
```
mightyiam: In each of these add/remove row/column click handlers, `false` is always returned (unless there is an error in the method).

Returning `false` doesn't seem to have anything to do with the methods.
So I've plucked it out of the methods and put it unconditionally at the end of the each handler.

src/wymeditor/plugins/table/jquery.wymeditor.table.js
```diff
@@ -138,22 +138,38 @@ TableEditor.prototype.bindEvents = function () {
...
+        var cont = wym.selectedContainer();
+        if (cont) {
+            tableEditor.addRow(cont);
+        }
```
mightyiam: I've also added a conditional, here. It is important, for when the call to `selectedContainer` during some selections returns `false`.

That currently produces an error inside the add/remove row/column methods.

This fixes it.

src/wymeditor/plugins/table/jquery.wymeditor.table.js
```diff
@@ -394,13 +410,13 @@ TableEditor.prototype.mergeRow = function () {
     // Just use the td and th nodes
     cells = jQuery(nodes).filter('td,th');
     if (cells.length === 0) {
-        return false;
+        return;
```
mightyiam: So there is no need to return `false` here or at the end. Ditto for all the manipulation methods here.

mightyiam:
Writing unit tests for this would be a waste of time and energy.

src/wymeditor/plugins/table/jquery.wymeditor.table.js
winhamwr: :+1:

winhamwr:
I'd prefer at least one unit test for this, but if you're OK with manual testing, that's your call. I'm also feeling that the unittesting trade-off on WYMeditor is different than it was 3 years ago.

src/wymeditor/plugins/table/jquery.wymeditor.table.js
```diff
@@ -138,22 +138,38 @@ TableEditor.prototype.bindEvents = function () {
...
+        if (cont) {
+            tableEditor.addRow(cont);
+        }
+        return false;
     });
     jQuery(wym._box).find(
         tableEditor._options.sRemoveRowButtonSelector
     ).click(function () {
-        return tableEditor.removeRow(wym.selectedContainer());
+        var cont = wym.selectedContainer();
```
winhamwr: What does `selectedContainer()` return if the collection is across multiple blocks? I think at least some of the table methods handle a multiple selection, so we need to make sure we're not breaking that.

mightyiam: It returns `false`.

`selectedContainer` is not the exact most correct thing to use in, row removal, for one.

Because, for a selection like this one:
screenshot_2015-11-27_18-11-50
It returns `false`.

So, in `master`, this throws and in this branch it aborts row removal.

So, if you'd like, you can report this and I'll have to come up with some abstraction for such a selection, like `getSelectionFirstCommonAncestor` or something.

In any case, this is a pareto improvement, because aborting is better than throwing.

mightyiam:
Thank you!

PR #748, opened 2015-11-22
Reposition caret when removing row/column that contains selection

mightyiam:


src/test/unit/plugin_tests/table_test.js
```diff
@@ -905,120 +905,85 @@ test("Deleting all columns removes table", function () {
...
             strictEqual(
-                wymeditor.hasSelection(),
-                false,
-                "No selection"
+                wymeditor.selectedContainer(),
+                prevTd,
+                "selected container"
             );
-        },
-        expectedResultHtml: removedRow3Html
```
mightyiam: No good to check the result HTML because this isn't what this test is about.

Same for all of these tests.

mightyiam:
Manual testing produced fab results.

src/test/unit/plugin_tests/table_test.js
winhamwr: I liked also checking the HTML because it ensures that we didn't screw up the test setup and the row was actually removed. There's a purity for pragmatism trade-off, there, but your call.

winhamwr:
:+1: 
Looks good!

src/test/unit/plugin_tests/table_test.js
mightyiam: Do you think there are not unit tests for the manipulations themselves?

Or, do you suggest that this would test that upon this selection the row is also removed?

mightyiam:
Thank you!

PR #749, opened 2015-11-29
Register any modification after selection change

mightyiam:


mightyiam:
Works lovely in all browsers.

package.json
```diff
@@ -26,7 +26,7 @@
   "devDependencies": {
     "lodash.times": "^3.0.1",
     "combokeys": "^2.4.3",
-    "edited": "^1.1.1",
+    "edited": "^1.1.2",
```
mightyiam: I love fixing things upstream.

winhamwr: Minor quibble, but I would say that this is more of a new feature in `edited` than a bugfix, so I probably would have bumped it to `1.2.0`.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ WYMeditor.
 * [#764](https://github.com/wymeditor/wymeditor/pull/764)
   TablePlugin row/column remove:
   refrain from error when no selectedContainer
+* [#749](Register any modification after selection change)
```
winhamwr: How about:

If a user modifies their selection via arrow keys or clicking, a new undo/redo savepoint is now automatically created. Previously, typing, moving selection, typing some more and then clicking undo would remove both typing actions. Now, only the second bit of typing will be undone.

package.json
mightyiam: In the scope of `edited`, it does seem like a feature, now that you say it.

CHANGELOG.md
mightyiam: I made it better.

mightyiam:
Thank you!

PR #766, opened 2015-11-29
Travis container based infrastructure

mightyiam:
https://docs.travis-ci.com/user/migrating-from-legacy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 before_script: npm install -g grunt-cli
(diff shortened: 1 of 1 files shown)


winhamwr:
With a sample size of 1, looks like the container-based build was ~5% faster than the last build not using it. When I merged in, the new build started in master almost instantly, which was cool. Looks like it started with 5 of the builds in parallel, too, which is also nice (not sure what the parallelism looked like previously).

mightyiam:
Thank you!

PR #767, opened 2015-11-29
set undoRedo.hasUnregisteredModification to null on init

mightyiam:


src/wymeditor/editor/undo-redo.js
```diff
@@ -38,6 +38,8 @@ WYMeditor.UndoRedo = function (wym) {
             return false;
         }
     );
+
+    undoRedo.hasUnregisteredModification = null;
```
mightyiam: Good for debugging.

PR #765, opened 2015-12-17
Resize handle is registered in undo history

mightyiam:
It seems that way.

### DIFF
--- a/src/test/unit/specific_feature_tests/image-resize-handle.js
+++ b/src/test/unit/specific_feature_tests/image-resize-handle.js
@@ -1,5 +1,7 @@
 /* jshint maxlen: 100 */
 /* global
+    expect,
+    sinon,
     manipulationTestHelper,
     prepareUnitTestModule,
     skipKeyboardShortcutTests,
@@ -126,4 +128,18 @@ test("resize handle with no image is hidden on `mousemove`", function () {
             strictEqual(wymeditor.$body().find('.wym-resize-handle').css('display'), 'none');
         }
     });
+});
+
+test("resize handle detached before adding history", function () {
+    expect(1);
+    var wymeditor = jQuery.wymeditors(0);
+    var spy = sinon.spy(wymeditor.ih, '_detachResizeHandle');
+
+    wymeditor.undoRedo.reset();
+    wymeditor.html('<p><img src="' + IMG_SRC + '" /></p>');
+    wymeditor.$body().find('img')
+        .mousemove();
+    wymeditor.undoRedo._add();
+
+    strictEqual(spy.calledOnce, true);
 });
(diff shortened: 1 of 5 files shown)


src/test/unit/specific_feature_tests/image-resize-handle.js
```diff
@@ -127,3 +129,17 @@ test("resize handle with no image is hidden on `mousemove`", function () {
...
+    var wymeditor = jQuery.wymeditors(0);
+    var spy = sinon.spy(wymeditor.ih, '_detachResizeHandle');
+
+    wymeditor.undoRedo.reset();
+    wymeditor.html('<p><img src="' + IMG_SRC + '" /></p>');
+    wymeditor.$body().find('img')
+        .mousemove();
+    wymeditor.undoRedo._add();
+
+    strictEqual(spy.calledOnce, true);
```
michaelgzoller: You're testing that _detachResizeHandle is called once, which doesn't seem the same as testing that it's not added to the undo/redo stack to me.

michaelgzoller:
I'm also not sure what the existing vs expected behavior is here. Are you saying that if you resize something, then click undo, the image will revert to its previous size?

michaelgzoller:
To make this easier to review, can you please add a better description of the bug and the expected/desired behavior?

src/wymeditor/editor/undo-redo.js
```diff
@@ -86,6 +86,12 @@ WYMeditor.UndoRedo.prototype._add = function () {
         return;
     }
 
+    if (wym.ih._isResizeHandleAttached()) {
+        // so that the resize handle
+        // does not end up as content in the history
+        wym.ih._detachResizeHandle();
```
winhamwr: I really wish we had called this variable `imageHandler` instead of `ih`. I had to grep through source to figure out what was happening.

winhamwr: Does this mean that if someone clicks to resize the handler, then clicks somewhere else, then resize handler will poof? Is that OK?

winhamwr:
Tested it out and it's OK that the handler poofs when undo is registered. Any undo event should already be poofing the handler.

:+1:

src/test/unit/specific_feature_tests/image-resize-handle.js
mightyiam: You're right, @michaelgzoller, it is not the same thing.
Also, it would be better to test that `_detachResizeHandle` is called before `_add` is called. And I don't do that, either. This seems to be the correct amount of love this old editor should get.

mightyiam:
Image resizes are not currently saved in undo history, actually. It is not clear where this operation should be saved and there's some complexity to it.

mightyiam:
Thank you!

PR #745, opened 2015-12-19
Throws when attempting to change multiple root containers

mightyiam:
[In `setRootContainer`](https://github.com/wymeditor/wymeditor/blob/9311f8035bafc0f412bf47a98a0aa31d8df9aa56/src/wymeditor/editor/base.js#L1049) there is no check whether `selectedContainer` returns `false`, which is not a valid value for `findUp`.

### DIFF
--- a/src/wymeditor/editor/base.js
+++ b/src/wymeditor/editor/base.js
@@ -1028,7 +1028,7 @@ WYMeditor.editor.prototype.getRootContainer = function () {
 */
 WYMeditor.editor.prototype.setRootContainer = function (sType) {
     var wym = this,
-        container = null,
+        selectedContainer = null,
         aTypes,
         newNode,
         blockquote,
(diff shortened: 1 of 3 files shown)


src/wymeditor/editor/base.js
```diff
@@ -1037,9 +1037,12 @@ WYMeditor.editor.prototype.setRootContainer = function (sType) {
         firstNode,
         x;
 
-    if (sType.toLowerCase() === WYMeditor.TH) {
-        container = wym.selectedContainer();
+    container = wym.selectedContainer();
```
winhamwr: nitpick. Could we call this variable `selectedContainer`? I know it was already called `container`, initially.

winhamwr:
:+1: 

Hooray for not throwing exceptions :)

mightyiam:
Thank you!

Issue #681, opened 2015-03-03
undoRedo.registerModification on preModification


mightyiam:
Modification performing actions which call `registerModification` at their end should also call a yet unexistent `preModification`.

`preModification` will call `registerModification` if `wym.undoRedo.hasUnregisteredEdits`.

That way we don't lose those unregistered edits.

PR #768, opened 2015-12-27
`undoRedo.registerModification` should be called on UI interaction

mightyiam:
`undoRedo.registerModification` should be called only on the level of UI interaction.

For example, it should be called from the handler of a toolbar button.

It should not be called from the manipulation functions.

Why?

Because a "modification", that can be undone, from the user's perspective, is a modification that he himself triggered.

So, whatever was the result of him pressing a button should be undone.

A button could potentially call for several consecutive modifications.

In that case, we certainly don't want several history points, do we?

This is why the calls to `undoRedo.registerModificaion`  must be taken out of the modification functions/methods and put in the UI actions handlers.

And this should happen before #681, so that we don't add another thing to the modification functions.

### DIFF
--- a/src/test/unit/plugin_tests/table_test.js
+++ b/src/test/unit/plugin_tests/table_test.js
@@ -713,8 +713,7 @@ test("no-op on non-table elements", function () {
             var actionElement = wymeditor.$body().find('#p_1')[0];
             wymeditor.tableEditor.addColumn(actionElement);
         },
-        expectedResultHtml: basicWithPHtml,
-        testUndoRedo: true
+        expectedResultHtml: basicWithPHtml
     });
 
     manipulationTestHelper({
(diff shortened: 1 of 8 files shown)


mightyiam:
There's something really wrong with some of our manipulation functions. Many of them have within them too much scope.

They do two things which should be separate:
* Determining whether a manipulation should occur, depending on the state of the selection and the content.
* Making that manipulation.

The former should reside in the handler of the UI action (click on toolbar item, for example).
The latter should be a well scoped DOM manipulating function.

This way, the former will be able to call some structure checking function(s) and determine whether the user wishes a manipulation (the latter) or more and call those.

I really don't think I'll ever fix all of these. It is great to learn, though.

mightyiam:
Next step is to call `registerModification` conditionally at the end of each UI action handler.

The condition is whether there was a modification or not. This relies on each of the manipulation functions to correctly return a boolean regarding that.

So each of the manipulation functions should be unit tested for that return value.

But like I wrote just above, not all of the manipulation functions should even have that logic of whether to perform a manipulation or not. Of most of them, if not all, I expect to be dumb about it and just do a manipulation.

PR #775, opened 2016-01-18
Adding an underscore to the console warning.

mpj17:
When a skin that is specified in wym._options cannot be found then a warning is displayed in the console. The warning had a missing underscore (_), so it caused an error.


winhamwr:
Thanks, Michael!

PR #776, opened 2016-01-26
Fix inserting images into classless spans

caffodian:
Inserting images into spans doesn't work, because it calls `toLowerCase()` on the null class attribute.


caffodian:
Hm.  That's not quite right.

PR #777, opened 2016-01-27
Fix error when inserting image into a span inside a list

caffodian:
There is a bug in the webkit span fixer code where when it is called on a span without a class, an error occurs.  The most common place we see this is when inserting images into spans within lists.  (Span alone causes the span to split, instead)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ WYMeditor.
 ### Bug Fixes
 
 * [#777](https://github.com/wymeditor/wymeditor/pull/777)
-  Fixed error when inserting into span nested inside list.
+  Fixed script error when inserting into span nested inside list.
 * [#761](https://github.com/wymeditor/wymeditor/pull/761)
   Image resize handle below end of iframe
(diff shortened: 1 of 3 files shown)


winhamwr:
> a bug in the webkit span fixer code

Just to be clear, this bug exists with classless spans in any browser, right?

winhamwr:
This needs an item in the [changelog](https://github.com/wymeditor/wymeditor/blob/master/CHANGELOG.md). After that, it looks good.

Up for [making a release](https://wymeditor.readthedocs.org/en/v1.1.1/wymeditor_development/making_releases.html) or would you prefer to wait until the other insertion bug is fixed?

caffodian:
> Just to be clear, this bug exists with classless spans in any browser, right?

It happens in Chrome, at least.  The class attr is undefined and then `toLowerCase` gets called on it.

> This needs an item in the changelog. After that, it looks good.
>
> Up for making a release or would you prefer to wait until the other insertion bug is fixed?

Will do, still figuring out how to contribute.  I haven't tracked down our other insertion bug yet, but I suspect that this fix also covers doing nearly anything inside a span (as long as the span doesn't get split by the command)

PR #782, opened 2017-01-23
Translatable themes 2

gyuris:
@winhamwr, @mightyiam I use wymeditor in Drupal. Recently I [swithed in Drupal](https://www.drupal.org/node/2273635) to wymeditor 1.1.1 and I realized that my original [PR](https://github.com/wymeditor/wymeditor/pull/338) wasn't merged in 2012. So I made a quick new PR, compatible with 1.1.1.

Has you intent to merge this?
I tested yet only in Gecko and Chrome. If you have intent to merge this patch, I can work on it.


mightyiam:
I don't have much incentive to support this project. I see it as a bit of a relic. I would suggest migrating to more modern editors.

